PR #4, opened 2024-09-23
Implement DNS resolver

clintonpi:
Motivation:

The GRPCNIOTransportCore module needs a DNS resolver, as currently, IP addresses have to be passed in manually.

Modifications:

- Add a new type `DNSResolver` with a method `resolve(host:port:)` that calls the `getaddrinfo` `libc` function to resolve a hostname and port number to a list of socket addresses. `resolve(host:port:)` is non-blocking and asynchronous.

Result:

The GRPCHTTP2Core module will have a DNS resolver.

---

_This PR re-opens https://github.com/grpc/grpc-swift/pull/2060._

### DIFF
--- a/Sources/GRPCNIOTransportCore/Client/Resolver/DNSResolver.swift
+++ b/Sources/GRPCNIOTransportCore/Client/Resolver/DNSResolver.swift
@@ -1,0 +1,255 @@
+/*
+ * Copyright 2024, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+private import Dispatch
+
+#if canImport(Darwin)
+package import Darwin
+#elseif canImport(Glibc)
+package import Glibc
+#elseif canImport(Musl)
+package import Musl
+#else
+#error("The GRPCNIOTransportCore module was unable to identify your C library.")
+#endif
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+/// An asynchronous non-blocking DNS resolver built on top of the libc `getaddrinfo` function.
+package enum DNSResolver {
+  private static let dispatchQueue = DispatchQueue(
+    label: "io.grpc.DNSResolver"
+  )
+
+  /// Resolves a hostname and port number to a list of socket addresses. This method is non-blocking.
+  package static func resolve(host: String, port: Int) async throws -> [SocketAddress] {
+    try Task.checkCancellation()
+
+    return try await withCheckedThrowingContinuation { continuation in
+      Self.dispatchQueue.async {
+        do {
+          let result = try Self.resolveBlocking(host: host, port: port)
+          continuation.resume(returning: result)
+        } catch {
+          continuation.resume(throwing: error)
+        }
+      }
+    }
+  }
+
+  /// Resolves a hostname and port number to a list of socket addresses.
+  ///
+  /// Calls to `getaddrinfo` are blocking and this method calls `getaddrinfo` directly. Hence, this method is also blocking.
+  private static func resolveBlocking(host: String, port: Int) throws -> [SocketAddress] {
+    var result: UnsafeMutablePointer<addrinfo>?
+    defer {
+      if let result {
+        // Release memory allocated by a successful call to getaddrinfo
+        freeaddrinfo(result)
+      }
+    }
+
+    var hints = addrinfo()
+    hints.ai_socktype = SOCK_STREAM
+    hints.ai_protocol = IPPROTO_TCP
+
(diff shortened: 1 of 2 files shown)


Sources/GRPCNIOTransportCore/Client/Resolver/DNSResolver.swift
```diff
@@ -0,0 +1,255 @@
...
+ */
+
+private import Dispatch
+
+#if canImport(Darwin)
+package import Darwin
+#elseif canImport(Glibc)
+package import Glibc
+#elseif canImport(Musl)
+package import Musl
```
glbrntt, line 24: I think these should be `private`, not sure it makes sense for types from lib C to make it beyond the scope of this file.

Tests/GRPCNIOTransportCoreTests/Client/Resolver/DNSResolverTests.swift
```diff
@@ -0,0 +1,37 @@
...
+
+    let expectedIPv4Result: [SocketAddress] = [.ipv4(host: "127.0.0.1", port: 80)]
+    let expectedIPv6Result: [SocketAddress] = [.ipv6(host: "::1", port: 80)]
+
+    let ipv4Result = try await DNSResolver.resolve(host: "127.0.0.1", port: 80)
+    let ipv6Result = try await DNSResolver.resolve(host: "::1", port: 80)
+
+    #expect(ipv4Result == expectedIPv4Result)
+    #expect(ipv6Result == expectedIPv6Result)
+  }
```
glbrntt, line 36: This can be a paremeterized test, can be something like this:

```suggestion
  @Test(
    "Resolve hostname", 
    arguments: [
      ("127.0.0.1", .ipv4(host: "127.0.0.1", port: 80)),
      ("::1", .ipv6(host: "::1", port: 80)),
    ] as [(String, SocketAddress)].self 
  )
  func resolve(host: String, expected: SocketAddress) async throws {
    // Note: This test checks the IPv4 and IPv6 addresses separately (instead of
    // `DNSResolver.resolve(host: "localhost", port: 80)`) because the ordering of the resulting
    // list containing both IP address versions can be different.
    
    let result = try await DNSResolver.resolve(host: host, port: 80)
    #expect(result == expected)
  }
```

glbrntt (changes requested):
It looks like the CI is failing because some symbols are missing, I think we need to add a C shim for Linux to get hold of these, let me do that in a separate PR.

glbrntt:
> It looks like the CI is failing because some symbols are missing, I think we need to add a C shim for Linux to get hold of these, let me do that in a separate PR.

Okay, multiple problems here:

- `EAI_ADDRFAMILY` and `EAI_NODATA` aren't available on Linux without including `<netdb.h>` on _and_ defining the `_GNU_SOURCE` feature test macro. We can work around this by adding a C shim library. All good.
- `EAI_BADHINTS` and `EAI_PROTOCOL` don't exist in Linux. We _could_ declare them in a platform specific way.

I have an alternative proposal to adding a C-shim module and doing all of the platform shenanigans: just call `gai_strerror` with the error code instead. It's job is to provide a string description of the error, and that's ultimately what we want in our error:

```swift
  package struct GetAddrInfoError: Error, Hashable, CustomStringConvertible {
    package let description: String

    package init(code: CInt) {
      if let cString = gai_strerror(code) {
        self.description = String(cString: cString)
      } else {
        self.description = "Unknown error: \(code)"
      }
    }
  }

```

Sources/GRPCNIOTransportCore/Client/Resolver/DNSResolver.swift
```diff
@@ -0,0 +1,255 @@
...
+    }
+  }
+}
+
+@available(macOS 10.15, iOS 13.0, watchOS 6.0, tvOS 13.0, *)
+extension DNSResolver {
+  /// `Error` that may be thrown based on the system error encountered by `inet_ntop`.
+  package enum InetNetworkToPresentationError: Error, Hashable {
+    case systemError(errno: errno_t)
+  }
```
glbrntt, line 224: Let's just make this a `struct` which stores the `errno` as a `CInt`

Sources/GRPCNIOTransportCore/Client/Resolver/DNSResolver.swift
```diff
@@ -0,0 +1,255 @@
...
+    defer {
+      if let result {
+        // Release memory allocated by a successful call to getaddrinfo
+        freeaddrinfo(result)
+      }
+    }
+
+    var hints = addrinfo()
+    hints.ai_socktype = SOCK_STREAM
+    hints.ai_protocol = IPPROTO_TCP
```
glbrntt, line 66: We need to convert to `CInt` here (which is expected by `ai_protocol`) as `IPPROTO_TCP` isn't always defined as a `CInt`:

```suggestion
    hints.ai_protocol = CInt(IPPROTO_TCP)
```

Tests/GRPCNIOTransportCoreTests/Client/Resolver/DNSResolverTests.swift
```diff
@@ -0,0 +1,39 @@
...
+  )
+  func resolve(host: String, expected: SocketAddress) async throws {
+    // Note: This test checks the IPv4 and IPv6 addresses separately (instead of
+    // `DNSResolver.resolve(host: "localhost", port: 80)`) because the ordering of the resulting
+    // list containing both IP address versions can be different.
+
+    let result = try await DNSResolver.resolve(host: host, port: 80)
+
+    #expect(result.count == 1)
+    #expect(result[0] == expected)
```
glbrntt, line 37:
```suggestion
    #expect(result == [expected])
```

glbrntt (approved):
Great, thanks @clintonpi

PR #5, opened 2024-09-23
Add a DNS NameResolver

glbrntt:
Motivation:

Many users will rely on DNS to resolve the IP addresses of servers to connect to, we should therefore provide a DNS name resolver.

Modifications:

- Add a DNS name resolver factory capable of resolving IP addresses
- Add the resolver to the registry defaults

Result:

Can resolve DNS targets

### DIFF
--- a/Sources/GRPCNIOTransportCore/Client/Resolver/NameResolver+DNS.swift
+++ b/Sources/GRPCNIOTransportCore/Client/Resolver/NameResolver+DNS.swift
@@ -1,0 +1,125 @@
+/*
+ * Copyright 2024, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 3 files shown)


Sources/GRPCNIOTransportCore/Client/Resolver/NameResolver+DNS.swift
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    func next() async throws -> NameResolutionResult? {
+      return try await self.next(isolation: nil)
+    }
+
+    func next(
+      isolation actor: isolated (any Actor)?
+    ) async throws(any Error) -> NameResolutionResult? {
+      return try await self.resolver.resolve()
```
gjcairo, line 122: Should `resolve()` take the `isolation` param too?

gjcairo (commented):
LGTM, left a comment about isolation

Sources/GRPCNIOTransportCore/Client/Resolver/NameResolver+DNS.swift
glbrntt: Yes I believe so, good catch

gjcairo (approved):

PR #11, opened 2024-10-06
Surface transient channel errors

gjcairo:
This PR better surfaces transient errors happening deeper in the channel to provide more information to the user.

It depends on https://github.com/grpc/grpc-swift/pull/2083 and https://github.com/grpc/grpc-swift/pull/2084.

### DIFF
--- a/Sources/GRPCNIOTransportCore/Client/Connection/Connection.swift
+++ b/Sources/GRPCNIOTransportCore/Client/Connection/Connection.swift
@@ -68,7 +68,7 @@ package final class Connection: Sendable {
     /// Closed because the remote peer initiate shutdown (i.e. sent a GOAWAY frame).
     case remote
     /// Closed because the connection encountered an unexpected error.
-    case error(any Error, wasIdle: Bool)
+    case error(RPCError, wasIdle: Bool)
   }
 
   /// Inputs to the 'run' method.
@@ -236,6 +236,7 @@ package final class Connection: Sendable {
     // This state is tracked here so that if the connection events sequence finishes and the
     // connection never became ready then the connection can report that the connect failed.
     var isReady = false
+    var unexpectedCloseError: (any Error)?
 
     func makeNeverReadyError(cause: (any Error)?) -> RPCError {
       return RPCError(
@@ -265,10 +266,15 @@ package final class Connection: Sendable {
             // The connection will close at some point soon, yield a notification for this
             // because the close might not be imminent and this could result in address resolution.
             self.event.continuation.yield(.goingAway(errorCode, reason))
-          case .idle, .keepaliveExpired, .initiatedLocally, .unexpected:
+          case .idle, .keepaliveExpired, .initiatedLocally:
             // The connection will be closed imminently in these cases there's no need to do
             // anything.
             ()
+          case .unexpected(let error, _):
+            // The connection will be closed imminently in this case.
+            // We'll store the error that caused the unexpected closure so we
+            // can surface it.
+            unexpectedCloseError = error
           }
 
           // Take the reason with the highest precedence. A GOAWAY may be superseded by user
@@ -318,7 +324,7 @@ package final class Connection: Sendable {
         finalEvent = .closed(connectionCloseReason)
       } else {
         // The connection never became ready, this therefore counts as a failed connect attempt.
-        finalEvent = .connectFailed(makeNeverReadyError(cause: nil))
+        finalEvent = .connectFailed(makeNeverReadyError(cause: unexpectedCloseError))
       }
 
       // The connection events sequence has finished: the connection is now closed.
--- a/Sources/GRPCNIOTransportCore/Client/Connection/Connection.swift
+++ b/Sources/GRPCNIOTransportCore/Client/Connection/Connection.swift
@@ -130,13 +130,14 @@ package final class Connection: Sendable {
     func establishConnectionOrThrow() async throws(RPCError) -> HTTP2Connection {
       do {
         return try await self.http2Connector.establishConnection(to: self.address)
+      } catch let error as RPCError {
+        throw error
       } catch {
-        throw (error as? RPCError)
-          ?? RPCError(
-            code: .unavailable,
-            message: "Could not establish a connection to \(self.address).",
-            cause: error
-          )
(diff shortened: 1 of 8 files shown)


glbrntt (changes requested):
Can you split this into two PRs, one for surfacing the errors and one for adding the tests? Makes it easier to review and we get a clearer history.

Sources/GRPCNIOTransportCore/Client/Connection/GRPCChannel.swift
```diff
@@ -758,14 +758,30 @@ extension GRPCChannel.StateMachine {
...
           actions.resumeContinuations = ConnectivityStateChangeActions.ResumableContinuations(
             continuations: continuations,
-            result: .failure(RPCError(code: .unavailable, message: "channel isn't ready"))
+            result: .failure(
+              RPCError(
+                code: .unavailable,
+                message: "Channel isn't ready.",
+                cause: cause
+              )
+            )
```
glbrntt, line 774: IIUC the cause is already constrained to be an `RPCError` here – we should just use it rather than wrapping it.

Sources/GRPCNIOTransportCore/Client/Connection/LoadBalancers/RoundRobinLoadBalancer.swift
```diff
@@ -735,6 +735,10 @@ extension ConnectivityState {
   static func aggregate(_ states: some Collection<ConnectivityState>) -> ConnectivityState {
     // See https://github.com/grpc/grpc/blob/master/doc/load-balancing.md
 
+    if states.isEmpty {
+      return .shutdown
+    }
```
glbrntt, line 740: I think I need some convincing on why this is correct – it's not obvious why it is.

Sources/GRPCNIOTransportCore/Client/Connection/LoadBalancers/RoundRobinLoadBalancer.swift
```diff
@@ -750,12 +754,16 @@ extension ConnectivityState {
...
-      return .transientFailure
+    // Otherwise, if all subchannels are in state TRANSIENT_FAILURE, the channel's state is TRANSIENT_FAILURE.
+    // Pick one of the errors to surface, as we can't surface all of them.
+    var cause: RPCError!
+    for state in states {
+      guard case .transientFailure(let error) = state else {
+        return .shutdown
+      }
+      cause = error
     }
```
glbrntt, line 765: Can we exhaustively switch over the state here? I think it helps make the code more obviously correct:

```swift
var cause: RPCError?
for state in states {
  switch state {
  case .transientFailure(let error):
    // ...
  case .shutdown:
    return .shutdown
  case .idle, .connecting, .ready:
    fatalError("...")  // Unreachable as all are covered above.
  }
}
```

We also know that if `cause` is `nil` after this point it can only be so because  `states` is empty. IMO this makes it clearer that the returning `.shutdown` when there's no state is a fallback rather than behaviour we were implementing from some spec (as we have done for the other cases).

Sources/GRPCNIOTransportCore/Client/Connection/LoadBalancers/Subchannel.swift
```diff
@@ -282,16 +282,27 @@ extension Subchannel {
     }
   }
 
-  private func handleConnectFailedEvent(in group: inout DiscardingTaskGroup) {
+  private func handleConnectFailedEvent(in group: inout DiscardingTaskGroup, error: any Error) {
```
glbrntt, line 285: Rather than `any Error`, can we use `RPCError` and propagate it back to wherever the `connectFailed` event came from?

Sources/GRPCNIOTransportCore/Client/Connection/LoadBalancers/Subchannel.swift
```diff
@@ -646,9 +657,15 @@ extension Subchannel {
           self = .notConnected(NotConnected(from: state))
           onClosed = .emitIdle
 
-        case .keepaliveTimeout, .error(_, wasIdle: false):
+        case .keepaliveTimeout:
+          self = .notConnected(NotConnected(from: state))
+          onClosed = .emitTransientFailureAndReconnect(
+            cause: RPCError(code: .unavailable, message: "The keepalive timed out.")
```
glbrntt, line 663: nit: I think we can add a little more detail to the message here

```suggestion
            cause: RPCError(code: .unavailable, message: "The connection became unresponsive and was closed because the keepalive timeout fired.")
```

glbrntt (changes requested):
This mostly looks great, left a couple of suggestions on how we can improve it a little more.

Sources/GRPCNIOTransportCore/Client/Connection/LoadBalancers/Subchannel.swift
gjcairo: Okay I had to do some gymnastics to get this done - not sure if it's compiler errors or I'm misunderstanding something about how Result deals with typed throws, but it wasn't too straightforward 😅

Sources/GRPCNIOTransportCore/Client/Connection/LoadBalancers/RoundRobinLoadBalancer.swift
gjcairo: I got rid of this based on your comment below.

Sources/GRPCNIOTransportCore/Client/Connection/Connection.swift
```diff
@@ -127,9 +127,19 @@ package final class Connection: Sendable {
   /// This function returns when the connection has closed. You can observe connection events
   /// by consuming the ``events`` sequence.
   package func run() async {
-    let connectResult = await Result {
-      try await self.http2Connector.establishConnection(to: self.address)
+    func establishConnectionOrThrow() async throws(RPCError) -> HTTP2Connection {
```
glbrntt, line 130: These are the right gymnastics: the other way to do this is to be explicit with the types in the `catching:` closure:

```swift
await Result { () async throws(RPCError) -> HTTP2Connection in
  // ...
}
```

PR #12, opened 2024-10-08
Add E2E tests with TLS enabled

gjcairo:
This PR adds E2E tests for `NIOPosix`'s transport implementation with TLS enabled. It also adds some additional API to create TLS configs.

### DIFF
--- a/Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportTLSEnabledTests.swift
+++ b/Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportTLSEnabledTests.swift
@@ -1,0 +1,343 @@
+/*
+ * Copyright 2024, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import GRPCNIOTransportHTTP2Posix
+import Testing
+import X509
+import Crypto
+import SwiftASN1
+import Foundation
+import NIOSSL
+
+@Suite("HTTP/2 transport E2E tests with TLS enabled")
+struct HTTP2TransportTLSEnabledTests {
+  // - MARK: Test Utilities
+
+  // A combination of client and server transport kinds.
+  struct Transport: Sendable {
+    var server: ServerKind
+    var client: ClientKind
+
+    enum ClientKind: Sendable {
+      case posix(HTTP2ClientTransport.Posix.Config.TransportSecurity)
+    }
+
+    enum ServerKind: Sendable {
+      case posix(HTTP2ServerTransport.Posix.Config.TransportSecurity)
+    }
+  }
+
+  func executeUnaryRPCForEachTransportPair(
+    transportProvider: (TestSecurity) -> [Transport]
+  ) async throws {
+    let security = try TestSecurity()
+    for pair in transportProvider(security) {
+      try await withThrowingTaskGroup(of: Void.self) { group in
+        let (server, address) = try await self.runServer(
+          in: &group,
+          kind: pair.server
+        )
+
+        let target: any ResolvableTarget
+        if let ipv4 = address.ipv4 {
+          target = .ipv4(host: ipv4.host, port: ipv4.port)
+        } else if let ipv6 = address.ipv6 {
+          target = .ipv6(host: ipv6.host, port: ipv6.port)
+        } else if let uds = address.unixDomainSocket {
+          target = .unixDomainSocket(path: uds.path)
+        } else {
+          Issue.record("Unexpected address to connect to")
+          return
+        }
+
+        let client = try self.makeClient(
+          kind: pair.client,
+          target: target
+        )
+
+        group.addTask {
+          try await client.run()
+        }
+
+        let control = ControlClient(wrapping: client)
+        try await self.executeUnaryRPC(control: control, pair: pair)
+
+        server.beginGracefulShutdown()
+        client.beginGracefulShutdown()
+      }
+    }
+  }
+
+  private func runServer(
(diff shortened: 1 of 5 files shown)


Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportTLSEnabledTests.swift
```diff
@@ -0,0 +1,402 @@
...
+struct TestSecurity {
+  struct Server {
+    let certificate: [UInt8]
+    let key: [UInt8]
+  }
+
+  struct Client {
+    let certificate: [UInt8]
+    let key: [UInt8]
+  }
```
glbrntt, line 349: These types are the same: we can have a single struct called `KeyPair`/`CertificateKeyPair` or similar and just have one called client and one called server. I also think we should rename the `TestSecurity` to have a clearer name (i.e. pluralise whatever we call the this struct)

Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportTLSEnabledTests.swift
```diff
@@ -0,0 +1,402 @@
...
+
+  init() throws {
+    let server = try Self.createSelfSignedDERCertificateAndPrivateKey(name: "Server Certificate")
+    let client = try Self.createSelfSignedDERCertificateAndPrivateKey(name: "Client Certificate")
+
+    self.server = Server(certificate: server.cert, key: server.key)
+    self.client = Client(certificate: client.cert, key: client.key)
+  }
+
+  private static func createSelfSignedDERCertificateAndPrivateKey(
```
glbrntt, line 362: naming nit: factory functions should be "make" per https://www.swift.org/documentation/api-design-guidelines/

glbrntt (changes requested):
Good start, I think we can tidy up a lot of the test code though!

Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportTLSEnabledTests.swift
```diff
@@ -0,0 +1,402 @@
...
+    enum ClientKind: Sendable {
+      case posix(HTTP2ClientTransport.Posix.Config.TransportSecurity)
+    }
+
+    enum ServerKind: Sendable {
+      case posix(HTTP2ServerTransport.Posix.Config.TransportSecurity)
+    }
+  }
+
+  func executeUnaryRPCForEachTransportPair(
```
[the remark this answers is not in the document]
gjcairo: The reason why I did it like this was because the important thing in these tests actually **is** the transport security config: incorrect config can make the TLS handshake fail, and it has to be all properly set up for the handshake to work. Each test has differences in how the config is set up. I wanted _that_ to be the focus of the tests - without it, the rest of the test is not really important: we're doing a unary request just to test the handshake but really it doesn't matter what we do after the handshake is completed.

To be clear, I'm not against refactoring all this - I'm sure it can be improved - but I'm not sure what the best alternative is. I could move the config to the parametrised argument, but it can get pretty large (and once we add support for NIOTS it will be worse):
```
@Test(arguments: [.posix(
  .tls(
    .defaults {
      $0.trustRoots = .certificates([.bytes(security.server.certificate, format: .der)])
      $0.serverHostname = "localhost"
    }
  )
)
],
[.posix(
  .tls(
    .defaults(
      certificateChain: [.bytes(security.server.certificate, format: .der)],
      privateKey: .bytes(security.server.key, format: .der)
    )
  )
)]
```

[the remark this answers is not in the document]
glbrntt: I'm not suggesting we put all the config as a parameter for exactly the reason you pointed out, it's totally unwieldy! I'm suggesting we say the _type_ of transport we want to use and have the test call helpers which create the right transport config for you.

One of the things I don't like about these tests is that there's a lot of repetition in creating the config, and its not super obvious what the intent of the config is. There are only a few different skews of config:
- server, defaults
- server, mtls, with trust roots 
- server, mtls, with trust roots, and hostname
- client, defaults, and hostname
- client, mtls, with trust roots 
- client, mtls, with trust roots, and hostname

We can surely use this info to provide helpers that make these tests more readable and make the tls config more obvious?

Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportTLSEnabledTests.swift
```diff
@@ -0,0 +1,444 @@
...
+    try await self.withClientAndServer(
+      clientTransportSecurity: clientTransportConfig,
+      serverTransportSecurity: serverTransportConfig
+    ) { control in
+      await #expect(
+        throws: Never.self,
+        performing: {
+          try await self.executeUnaryRPC(control: control)
+        }
+      )
```
glbrntt, line 57: nit: prefer single-lines where possible and trailing-closure syntax (same comment applies to other places too):

```suggestion
      await #expect(throws: Never.self) {
        try await self.executeUnaryRPC(control: control)
      }
```

Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportTLSEnabledTests.swift
```diff
@@ -0,0 +1,444 @@
...
+      for: serverTransport,
+      certificateKeyPairs: certificateKeyPairs,
+      includeClientCertificateInTrustRoots: true
+    )
+
+    try await self.withClientAndServer(
+      clientTransportSecurity: clientTransportConfig,
+      serverTransportSecurity: serverTransportConfig
+    ) { control in
+      await #expect(
```
glbrntt, line 121: nit: multiple trailing closure syntax please ;)

Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportTLSEnabledTests.swift
```diff
@@ -0,0 +1,444 @@
...
+
+          return true
+        }
+      )
+    }
+  }
+
+  // - MARK: Test Utilities
+
+  enum TransportSecurity: Sendable {
```
glbrntt, line 211: `TransportKind` is probably a more appropriate name here, security would be TLS/plaintext

glbrntt (changes requested):
This is _much_ easier to follow than before! I left a couple of nits but otherwise it looks great.

glbrntt (approved):
LGTM assuming CI passes after https://github.com/grpc/grpc-swift-nio-transport/pull/13 is merged. Thanks Gus!

PR #13, opened 2024-10-10
Add support for RPC cancellation

glbrntt:
Motivation:

grpc-swift has support for RPC cancellation via a cancellation handler. It should be supported here as well.

Modifications:

- Modify the server stream handler so that it holds the cancellation handle and triggers cancellation appropriately.
- Modify the server stream handling to set a cancellation handle on the appropriate stream handler.
- Update tests

Result:

HTTP/2 server transport respects stream cancellation

### DIFF
--- a/Sources/GRPCNIOTransportCore/Server/CommonHTTP2ServerTransport.swift
+++ b/Sources/GRPCNIOTransportCore/Server/CommonHTTP2ServerTransport.swift
@@ -241,19 +241,35 @@ package final class CommonHTTP2ServerTransport<
         return
       }
 
-      let rpcStream = RPCStream(
-        descriptor: descriptor,
-        inbound: RPCAsyncSequence(wrapping: inbound),
-        outbound: RPCWriter.Closable(
-          wrapping: ServerConnection.Stream.Outbound(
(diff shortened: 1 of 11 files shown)


Sources/GRPCNIOTransportCore/Server/CommonHTTP2ServerTransport.swift
```diff
@@ -241,19 +241,35 @@ package final class CommonHTTP2ServerTransport<
...
+          // Sync is safe: this is on the right event loop.
+          let sync = stream.channel.pipeline.syncOperations
+
+          // Looking up the handler can fail if the channel is already closed, in which case
+          // cancel the handle directly.
+          do {
+            let handler = try sync.handler(type: GRPCServerStreamHandler.self)
+            handler.setCancellationHandle(handle)
+          } catch {
+            handle.cancel()
```
gjcairo, line 255: In the `catch` case, is it correct to keep executing and run the `streamHandler`? Wouldn't the channel be in an inconsistent/unexpected state (or closed)?

gjcairo (approved):
Great, thanks George!

PR #16, opened 2024-10-15
Close channel on errors

glbrntt:
Motivation:

NIO channel pipeline's should, in most cases, be closed when an error is caught. This is not currently the case for the client and server connection channels.

Modifications:

- Close on error unless it's safe to ignore the error (such as if it's a stream-level error)

Result:

Connections are closed if something unrecoverable happens

### DIFF
--- a/Sources/GRPCNIOTransportCore/Server/Connection/ServerConnectionManagementHandler.swift
+++ b/Sources/GRPCNIOTransportCore/Server/Connection/ServerConnectionManagementHandler.swift
@@ -331,7 +331,7 @@ package final class ServerConnectionManagementHandler: ChannelDuplexHandler {
       // This should be resolved in NIOHTTP2: https://github.com/apple/swift-nio-http2/issues/466
       //
       // Only close the connection if it's not already closing (as this is the state in which the
-      // error can be safely.
+      // error can be safely ignored).
       return !self.state.isClosing
 
     case is NIOHTTP2Errors.StreamError:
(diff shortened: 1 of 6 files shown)


Sources/GRPCNIOTransportCore/Server/Connection/ServerConnectionManagementHandler.swift
```diff
@@ -314,6 +314,37 @@ package final class ServerConnectionManagementHandler: ChannelDuplexHandler {
...
+    case is NIOHTTP2Errors.NoSuchStream:
+      // In most cases this represents incorrect client behaviour. However, NIOHTTP2 currently
+      // emits this error if a server receives a HEADERS frame for a new stream after having sent
+      // a GOAWAY frame. This can happen when a client opening a stream races with a server
+      // shutting down.
+      //
+      // This should be resolved in NIOHTTP2: https://github.com/apple/swift-nio-http2/issues/466
+      //
+      // Only close the connection if it's not already closing (as this is the state in which the
+      // error can be safely.
```
gjcairo, line 334: This sentence is unfinished

gjcairo (changes requested):
LGTM, good catch! Just a small nit about a comment.

Sources/GRPCNIOTransportCore/Server/Connection/ServerConnectionManagementHandler.swift
glbrntt, line 334:
```suggestion
      // error can be safely ignored).
```

gjcairo (approved):

PR #18, opened 2024-10-29
Add support for additional TLS options in NIOTS transport

gjcairo:
This PR adds support for a few additional TLS options in the NIOTS transport, to bring it in line with the NIOPosix implementation. In particular, it now supports custom trust roots, which means mTLS can be enabled when using NIOTS, and E2E testing is possible (because we can use self-signed certificates).

Some code has been refactored to be shared across transport implementations, and a small bug with how errors were being surfaced has been fixed.


Sources/GRPCNIOTransportHTTP2TransportServices/HTTP2ClientTransport+TransportServices.swift
```diff
@@ -327,6 +368,54 @@ extension NWProtocolTLS.Options {
         `protocol`
       )
     }
+
+    switch tlsConfig.trustRoots.wrapped {
+    case .certificates(let certificates):
+      let verifyQueue = DispatchQueue(label: "certificateVerificationQueue")
```
glbrntt, line 374: nit: should be prefixed with `io.grpc`, we can also drop "queue" from the name, it's a bit redundant here.

```suggestion
      let verifyQueue = DispatchQueue(label: "io.grpc.CertificateVerification")
```

Sources/GRPCNIOTransportHTTP2TransportServices/HTTP2ClientTransport+TransportServices.swift
```diff
@@ -327,6 +368,54 @@ extension NWProtocolTLS.Options {
...
+            let certificateBytes: Data
+            switch certificateSource.wrapped {
+            case .file(let path, .der):
+              certificateBytes = try Data(contentsOf: URL(filePath: path))
+
+            case .bytes(let bytes, .der):
+              certificateBytes = Data(bytes)
+
+            case .file(_, let format), .bytes(_, let format):
+              fatalError("Certificate format must be DER, but was \(format).")
```
glbrntt, line 390: Hmm – do we document this requirement somewhere?

Sources/GRPCNIOTransportHTTP2TransportServices/HTTP2ServerTransport+TransportServices.swift
```diff
@@ -264,6 +280,54 @@ extension NWProtocolTLS.Options {
...
+      }
+
+      sec_protocol_options_set_verify_block(
+        self.securityProtocolOptions,
+        verifyBlock,
+        verifyQueue
+      )
+
+    case .systemDefault:
+      ()
```
glbrntt, line 329: This is all copy-pasta from the client equivalent, can we avoid the duplication?

glbrntt (approved):
Looks great, thanks @gjcairo

glbrntt:
API breakage expected, the API was moved

```
1 breaking change detected in GRPCNIOTransportHTTP2Posix:
  💔 API breakage: enum TLSConfig has been removed
```

PR #21, opened 2024-11-04
Migrate to common GitHub Actions

rnro:
Migrate CI to use swiftlang / SwiftNIO common GitHub Actions.

### Motivation:

* Reduce duplication
* Centralise boilerplate changes when new Swift versions are picked up.
* Benefit from centralised work to add new linting / test
  infrastructure.

### Modifications:

Changes of note:
* Use soundness checks from swiftlang/github-workflows.
* Retain bespoke license-checking code for .swift files as the gRPC
  header style is very different to most templates.
* Remove scripts which are no longer needed.
* Move scripts to `scripts/` directory in-line with most other Swift on
  Server repositories.

### Result:

More test, linting, formatting coverage. More common CI with other Swift
on Server projects.


dev/protos/tests/interoperability/src/proto/grpc/testing/empty.proto
```diff
@@ -1,5 +1,4 @@
-
-// Copyright 2015 gRPC authors.
+// Copyright 2015, gRPC Authors All rights reserved.
```
glbrntt, line 1: The proto files comes from upstream repos (some can be updated via a script) so we shouldn't modify their copyright headers.

scripts/build-performance-tests.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
glbrntt, line 1: Can the scripts go back into `dev`, please? 

Originally we had both, but scripts ended up in `dev` and ultimately they're only used by the developers of the project anyway so `dev` felt more appropriate. The other benefit to just having `dev` is that it's a little tidier at the top-level.

dev/format.sh
```diff
@@ -1,92 +0,0 @@
-#!/bin/bash
```
glbrntt, line 1: Can we keep this script but use it solely for formatting? 

It's a much easier local workflow to run `./dev/format.sh` than to setup and run `act` -- especially so for new contributors.

.github/workflows/pull_request_label.yml
```diff
@@ -0,0 +1,18 @@
+name: PR label
```
glbrntt, line 1: Total nit, can this just be called "PR"? Or does the name need to be unique? I think there's enough context from the job name.

Screenshot 2024-11-04 at 16 41 36

glbrntt (changes requested):
Thanks for doing this @rnro – this is mostly great!

.github/workflows/pull_request_label.yml
rnro: No worries, no need for it to be unique, they can be uniquely referenced using the yaml file if needed. The listing now looks like this:
```
Stage  Job ID               Job name                      Workflow name  Workflow file           Events
0      unit-tests           Unit Tests                    Main           main.yml                push,schedule
0      integration-tests    Integration Tests             Main           main.yml                schedule,push
0      swift-license-check  Swift License Header Check    PR             pull_request.yml        pull_request
0      unit-tests           Unit Tests                    PR             pull_request.yml        pull_request
0      integration-tests    Integration Tests             PR             pull_request.yml        pull_request
0      cxx-interop          Cxx Interop                   PR             pull_request.yml        pull_request
0      soundness            Soundness                     PR             pull_request.yml        pull_request
0      semver-label-check   Semantic Version Label Check  PR             pull_request_label.yml  pull_request
```

dev/protos/tests/interoperability/src/proto/grpc/testing/empty.proto
rnro: I've backed out the changes and added the any `.proto` file under `./dev/protos/` to the ignore list.

glbrntt (approved):
Awesome, thanks @rnro

PR #31, opened 2024-11-22
Use new descriptor API

glbrntt:
Motivation:

GRPCCore changed some of the descriptor API.

Modifications:

- Use the new API

Result:

- Compiles again
- No warnings


glbrntt:
API breakage check fails because GRPCCore broke API.

PR #33, opened 2024-11-29
Propagate the authority pseudoheader

glbrntt:
Motivation:

We should be sending the ":authority" pseudoheader, somehow this was missed when doing the stream state machine.

The strategy for propagating the authority will be to:
- Use any override value set on the client transport, otherwise
- Use the value provided by a name resolver (if any), otherwise
- Derive a value from the target address.

Modifications:

- Add client config allowing the authority to be overridden
- Add authority to the name resolver interface
- Propagate the authority through to the stream state machine
- Add a percent encoder, as the authority should be percent encoded
- Remove hostname from the client TLS config and use the authority instead. This is generally a usability win as clients generally won't have to set this value themselves.

Result:

Authority is set automatically, and can be overridden if necessary.

### DIFF
--- a/Sources/GRPCNIOTransportCore/Client/Connection/Connection.swift
+++ b/Sources/GRPCNIOTransportCore/Client/Connection/Connection.swift
@@ -85,6 +85,10 @@ package final class Connection: Sendable {
   /// The address to connect to.
   private let address: SocketAddress
 
+  /// The server authority. If `nil`, a value will be computed based on the endpoint being
+  /// connected to.
+  private let authority: String?
+
   /// The default compression algorithm used for requests.
   private let defaultCompression: CompressionAlgorithm
 
@@ -109,11 +113,13 @@ package final class Connection: Sendable {
 
   package init(
     address: SocketAddress,
+    authority: String?,
     http2Connector: any HTTP2Connector,
     defaultCompression: CompressionAlgorithm,
     enabledCompression: CompressionAlgorithmSet
   ) {
     self.address = address
+    self.authority = authority
     self.defaultCompression = defaultCompression
     self.enabledCompression = enabledCompression
     self.http2Connector = http2Connector
@@ -129,7 +135,10 @@ package final class Connection: Sendable {
   package func run() async {
     func establishConnectionOrThrow() async throws(RPCError) -> HTTP2Connection {
       do {
-        return try await self.http2Connector.establishConnection(to: self.address)
+        return try await self.http2Connector.establishConnection(
+          to: self.address,
+          authority: self.authority ?? self.address.sniHostname
+        )
       } catch let error as RPCError {
         throw error
       } catch {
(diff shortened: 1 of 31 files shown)


Sources/GRPCNIOTransportCore/Client/Connection/Connection.swift
```diff
@@ -129,7 +135,10 @@ package final class Connection: Sendable {
   package func run() async {
     func establishConnectionOrThrow() async throws(RPCError) -> HTTP2Connection {
       do {
-        return try await self.http2Connector.establishConnection(to: self.address)
+        return try await self.http2Connector.establishConnection(
+          to: self.address,
+          authority: self.authority ?? self.address.sniHostname
```
rnro, line 140: Here you use `self.authority ?? self.address.sniHostname`, but further down the file you use `self.authority ?? self.address.authority` - is that intentional?

Sources/GRPCNIOTransportCore/Client/Connection/GRPCChannel.swift
```diff
@@ -82,6 +86,12 @@ package final class GRPCChannel: ClientTransport {
     self.input = AsyncStream.makeStream()
     self.connector = connector
 
+    if let authority = config.http2.authority ?? resolver.authority {
+      self.authority = PercentEncoding.encodeAuthority(authority)
```
rnro, line 90: Why does the percent-encoding happen at this level of the hierarchy rather than when it is first obtained?

rnro: If it remains sometimes percent-encoded and sometimes not when stored as a property on objects I think that should be documented.

Sources/GRPCNIOTransportCore/Internal/PercentEncoding.swift
```diff
@@ -0,0 +1,135 @@
...
+      Self.isAuthorityChar($0)
+    }
+  }
+
+  package static func encode(
+    _ input: String,
+    isValidCharacter: (UInt8) -> Bool
+  ) -> String {
+    var output: [UInt8] = []
+    output.reserveCapacity(input.utf8.count)
```
rnro, line 29: Do you think it's worth reserving by some amount greater than the count here, anticipating some conversions?

Sources/GRPCNIOTransportCore/Internal/PercentEncoding.swift
```diff
@@ -0,0 +1,135 @@
...
+
+    for char in input.utf8 {
+      if isValidCharacter(char) {
+        output.append(char)
+      } else {
+        output.append(UInt8(ascii: "%"))
```
rnro, line 35: Will the `UInt8` conversion of `%` be computed and stored at compile time? Is it worth doing that ourselves and hardcoding it e.g. `UInt8(0x25) // %`

rnro (approved):
Thanks!

glbrntt:
API breakage is expected and okay.

PR #36, opened 2024-12-03
Add remote peer info

glbrntt:
Motivation:

The core package added remote peer info to the server context. We should provide that info from connected channels.

Modification:

- Set the remote peer info
- Add tests

Result:

More insight into remote peer


glbrntt:
API breakage check fails to build as the API in the core package changed so the current version of `main` in this package won't compile (as it depends on `main` in the core package).

rnro (approved):

PR #37, opened 2024-12-04
Separate out the security config from the rest of the transport config

glbrntt:
Motivation:

Users must explicitly state whether they want TLS or plaintext. This config for this is currently nested within the rest of the config for the transport. In many cases this results in users writing `.defaults(transportSecurity: .tls(.defaults(...)))` where `.defaults(...)` often includes various TLS config. This is verbose and a little bit finicky to use.

Modifications:

- Seprate out the transport security from the rest of the config and have the client and server transport take a config _and_ a transport security config.
- This allows for APIs which compose better while also allowing for the transport config to just use a set of defaults.

Result:

Easier to use API.

### DIFF
--- a/Sources/GRPCNIOTransportHTTP2Posix/Config+TLS.swift
+++ b/Sources/GRPCNIOTransportHTTP2Posix/Config+TLS.swift
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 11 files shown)


gjcairo (approved):
Nice refactor, thanks George!

PR #45, opened 2024-12-09
Enable MemberImportVisibility check on all targets

rnro:
Enable MemberImportVisibility check on all targets. Use a standard string header and footer to bracket the new block for ease of updating in the future with scripts.


glbrntt:
@rnro nightly fails because of an import visibility check which should be fixed by https://github.com/grpc/grpc-swift-nio-transport/pull/47

I'll approve + merge etc., once the above has been merged (and any other fallout from setting the member import visibility)

glbrntt (approved):

PR #46, opened 2024-12-11
Fix TLS E2E test flakiness

gjcairo:
Some of the E2E tests with TLS enabled were flaky on Darwin.

This PR:
- Slightly modifies the way they are implemented to make them not-flaky
- Modifies the options passed to `SecPKCS12Import` to make sure the creation of the required `SecIdentity` happens in memory only and doesn't use the keychain
- Tolerates a broken pipe error when the server fails client cert validation during an mTLS handshake, as it's possible for Network.framework to expose this error instead of the actual SSL validation error on the client


glbrntt (approved):
Thanks a lot @gjcairo!

glbrntt:
@gjcairo FYI formatter is unhappy

PR #49, opened 2024-12-18
Reset go away state

### LINKED ISSUE #40: ”Channel isn't ready.” after client being ideal for a while
I see the following error message: ”14 UNAVAILABLE: Channel isn't ready.”. This happens for all client request after the process has been ideal for a while (sometimes < 5 minutes). Once this happens all requests fails. Restarting the process temporarily solves the issue.

I can reproduce this booth locally on macOS and when running containerized (Linux) in the cloud.

Let me know if this is a known issue already; if not, let me know if more details on how to reproduce this is needed.

Tested on [1.0.0-beta.1](https://github.com/grpc/grpc-swift-nio-transport/releases/tag/1.0.0-beta.1)-release.

glbrntt:
Motivation:

The pick first load balancer keeps track of whether the current subchannel is going away. This is done so that if there isn't yet a subchannel to roll over too it won't be offered up when the GRPCChannel asks for a subchannel. However, this state wasn't reset in enough places within the state machine.

Modifications:

- Reset the is-going-away state when updating the current subchannel and when the current subchannel becomes idle

Result:

Resolves #40

### DIFF
--- a/Tests/GRPCNIOTransportCoreTests/Client/Connection/LoadBalancers/PickFirstLoadBalancerTests.swift
+++ b/Tests/GRPCNIOTransportCoreTests/Client/Connection/LoadBalancers/PickFirstLoadBalancerTests.swift
@@ -295,6 +295,7 @@ final class PickFirstLoadBalancerTests: XCTestCase {
       case .connectivityStateChanged(.ready):
         switch idleCount.value {
         case 1:
+          XCTAssertNotNil(context.loadBalancer.pickSubchannel())
           // Must be connected to server 1, send a GOAWAY frame.
           let channel = context.servers[0].server.clients.first!
           let goAway = HTTP2Frame(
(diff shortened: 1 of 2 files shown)


Tests/GRPCNIOTransportCoreTests/Client/Connection/LoadBalancers/PickFirstLoadBalancerTests.swift
```diff
@@ -307,6 +308,7 @@ final class PickFirstLoadBalancerTests: XCTestCase {
           // Must only be connected to server 2 now.
           XCTAssertEqual(context.servers[0].server.clients.count, 0)
           XCTAssertEqual(context.servers[1].server.clients.count, 1)
+          XCTAssertNotNil(context.loadBalancer.pickSubchannel())
```
glbrntt, line 311: Without the fix this returns `nil` even though a connection is ready and available to use.

gjcairo (approved):
Good one, thanks George!

PR #50, opened 2024-12-20
Fix keepalive logic

gjcairo:
This PR fixes https://github.com/grpc/grpc-swift/issues/2095.

## Motivation

As per the gRPC specification, the server must keep track of pings from each client, and if they go over a threshold, we must send a GOAWAY frame and close the connection. We must reset the number of ping strikes every time the server writes a headers or data frame. However, there is a bug in the current keepalive implementation and we are not properly keeping track of when header/data frames are written, so we never reset the strikes, causing the server to always end up closing connections when keepalive pings are enabled.

There was also a second bug where the GOAWAY frame wasn't actually sent to the client because we were closing the connection straight away, and the packet never made it out.

## Modifications

This PR fixes a couple of bugs:
- It keeps track of the appropriate FrameStats as described above
- It delays the channel close after sending the GOAWAY packet by a tick to make sure it gets flushed and delivered to the client

## Results

Fewer bugs!

### DIFF
--- a/Sources/GRPCNIOTransportCore/Server/Connection/ServerConnectionManagementHandler.swift
+++ b/Sources/GRPCNIOTransportCore/Server/Connection/ServerConnectionManagementHandler.swift
@@ -609,6 +609,9 @@ extension ServerConnectionManagementHandler {
 
       context.write(self.wrapOutboundOut(goAway), promise: nil)
       self.maybeFlush(context: context)
+
+      // We must delay the channel close after sending the GOAWAY packet by a tick to make sure it
+      // gets flushed and delivered to the client before the connection is closed.
       let loopBound = NIOLoopBound(context, eventLoop: context.eventLoop)
       context.eventLoop.execute {
         loopBound.value.close(promise: nil)
--- a/Sources/GRPCNIOTransportCore/Server/Connection/ServerConnectionManagementHandler.swift
+++ b/Sources/GRPCNIOTransportCore/Server/Connection/ServerConnectionManagementHandler.swift
@@ -121,9 +121,9 @@ package final class ServerConnectionManagementHandler: ChannelDuplexHandler {
   }
 
   /// Stats about recently written frames. Used to determine whether to reset keep-alive state.
-  private var frameStats: FrameStats
+  package var frameStats: FrameStats
 
-  struct FrameStats {
+  package struct FrameStats {
     private(set) var didWriteHeadersOrData = false
 
     /// Mark that a HEADERS frame has been written.
@@ -609,7 +609,10 @@ extension ServerConnectionManagementHandler {
 
       context.write(self.wrapOutboundOut(goAway), promise: nil)
       self.maybeFlush(context: context)
-      context.close(promise: nil)
+      let loopBound = NIOLoopBound(context, eventLoop: context.eventLoop)
+      context.eventLoop.execute {
+        loopBound.value.close(promise: nil)
+      }
 
     case .sendAck:
       ()  // ACKs are sent by NIO's HTTP/2 handler, don't double ack.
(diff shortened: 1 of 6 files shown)


Sources/GRPCNIOTransportCore/Server/Connection/ServerConnectionManagementHandler.swift
```diff
@@ -609,7 +609,10 @@ extension ServerConnectionManagementHandler {
 
       context.write(self.wrapOutboundOut(goAway), promise: nil)
       self.maybeFlush(context: context)
-      context.close(promise: nil)
+      let loopBound = NIOLoopBound(context, eventLoop: context.eventLoop)
```
glbrntt, line 612: Can you add a short code comment here explaining why we delay the close

Tests/GRPCNIOTransportCoreTests/Server/GRPCServerStreamHandlerTests.swift
```diff
@@ -1045,6 +1080,60 @@ struct ServerStreamHandlerTests {
...
+    // Now write back headers and make sure FrameStats are updated accordingly:
+    // To do that, we first need to receive client's initial metadata...
+    let clientInitialMetadata: HPACKHeaders = [
+      GRPCHTTP2Keys.path.rawValue: "/SomeService/SomeMethod",
+      GRPCHTTP2Keys.scheme.rawValue: "http",
+      GRPCHTTP2Keys.method.rawValue: "POST",
+      GRPCHTTP2Keys.contentType.rawValue: "application/grpc",
+      GRPCHTTP2Keys.te.rawValue: "trailers",
+    ]
+    #expect(throws: Never.self) {
```
glbrntt, line 1107: Don't think we need to do this, just let the test throw if there's an error

Tests/GRPCNIOTransportCoreTests/Server/GRPCServerStreamHandlerTests.swift
```diff
@@ -1045,6 +1080,60 @@ struct ServerStreamHandlerTests {
...
+        HTTP2Frame.FramePayload.headers(.init(headers: clientInitialMetadata))
+      )
+    }
+
+    // Now we write back server's initial metadata...
+    let headers: HPACKHeaders = [
+      "some-custom-header": "some-custom-value"
+    ]
+    let serverInitialMetadata = RPCResponsePart.metadata(Metadata(headers: headers))
+    #expect(throws: Never.self) {
```
glbrntt, line 1118: Same here re: throwing

Tests/GRPCNIOTransportCoreTests/Server/GRPCServerStreamHandlerTests.swift
```diff
@@ -1045,6 +1080,60 @@ struct ServerStreamHandlerTests {
...
+      try channel.writeInbound(
+        HTTP2Frame.FramePayload.headers(.init(headers: clientInitialMetadata))
+      )
+    }
+
+    // Now we write back server's initial metadata...
+    let headers: HPACKHeaders = [
+      "some-custom-header": "some-custom-value"
+    ]
+    let serverInitialMetadata = RPCResponsePart.metadata(Metadata(headers: headers))
```
glbrntt, line 1117: We don't need to add extra headers here, just write the empty metadata and the state machine will add the appropriate http headers
```suggestion
    let serverInitialMetadata = RPCResponsePart.metadata([:])
```

glbrntt (changes requested):
Couple of small nits but looks great otherwise, thanks Gus!

glbrntt (approved):

PR #55, opened 2025-01-14
Adopt new bag-of-bytes protocol

glbrntt:
Motivation:

The core package added a bag-of-bytes protocol which transports must be generic over. We need to adopt those changes here.

Modifications:

- Add a `GRPCNIOTransportBytes` type which wraps `ByteBuffer` and implements the `GRPCContiguousBytes` protocol.
- Use this as the bytes type for client/server transports.
- Update the appropriate state machines to deal in terms of `ByteBuffer`
- Update the compressor/decompressor to deal in terms of `ByteBuffer`
- Update tests

Result:

Avoid unconditonal copying of messages to/from `[UInt8]`/`ByteBuffer`


glbrntt:
This depends on and won't compile until https://github.com/grpc/grpc-swift/pull/2155 is merged.

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ClientTransport+Posix.swift
```diff
@@ -105,7 +107,7 @@ extension HTTP2ClientTransport {
       self.channel.retryThrottle
     }
 
-    public func connect() async {
+    public func connect() async throws {
```
gjcairo, line 110: Is it a regular Swift convention to keep the `throws` from the protocol signature if we don't actually throw in the implementation?

glbrntt: I'm not sure if there is a convention. The `protocol` is `throws` so adding it back isn't an issue and leaves us scope in the future in case the implementation actually needs to become throwing.

Sources/GRPCNIOTransportCore/GRPCNIOTransportBytes.swift
```diff
@@ -0,0 +1,65 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+public import GRPCCore
+public import NIOCore
+
+/// The contiguous bytes type used by the gRPC's NIO transport.
+public struct GRPCNIOTransportBytes: GRPCContiguousBytes, Hashable, Sendable {
```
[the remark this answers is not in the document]
gjcairo: Right - I didn't consider this because this package is already so tied to NIO (and I find it hard to believe we'd replace ByteBuffer on NIO anytime soon 😄). But that's fair.

gjcairo:
Sorry, mistakenly clicked the not a draft anymore button

gjcairo (approved):

PR #56, opened 2025-01-15
Adopt `ClientContext` changes

gjcairo:
This PR adopts the changes introduced in https://github.com/grpc/grpc-swift/pull/2158, requiring client transports to provide a `ClientContext` alongside the stream.

### DIFF
--- a/Sources/GRPCNIOTransportCore/Client/Connection/Connection.swift
+++ b/Sources/GRPCNIOTransportCore/Client/Connection/Connection.swift
@@ -202,10 +202,10 @@ package final class Connection: Sendable {
     descriptor: MethodDescriptor,
     options: CallOptions
   ) async throws -> Stream {
-    let (multiplexer, scheme) = try self.state.withLock { state in
+    let (multiplexer, scheme, remotePeer, localPeer) = try self.state.withLock { state in
       switch state {
       case .connected(let connected):
-        return (connected.multiplexer, connected.scheme)
+        return (connected.multiplexer, connected.scheme, connected.remotePeer, connected.localPeer)
       case .notConnected, .closing, .closed:
         throw RPCError(code: .unavailable, message: "subchannel isn't ready")
       }
@@ -246,7 +246,13 @@ package final class Connection: Sendable {
         }
       }
 
-      return Stream(wrapping: stream, descriptor: descriptor)
+      let context = ClientContext(
+        descriptor: descriptor,
+        remotePeer: remotePeer,
+        localPeer: localPeer
+      )
+
+      return Stream(wrapping: stream, context: context)
     } catch {
       throw RPCError(code: .unavailable, message: "subchannel is unavailable", cause: error)
     }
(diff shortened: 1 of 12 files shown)


Sources/GRPCNIOTransportCore/Client/Connection/Connection.swift
```diff
@@ -421,6 +421,14 @@ extension Connection {
 
     private let http2Stream: NIOAsyncChannel<RPCResponsePart, RPCRequestPart>
 
+    var peerInfo: String {
```
glbrntt, line 424: I think we should call this `remoteInfo` to mirror its client counterpart

Sources/GRPCNIOTransportCore/Internal/Channel+AddressInfo.swift
```diff
@@ -0,0 +1,73 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import NIOCore
+
+extension Channel {
+  func getRemoteAddressInfo() -> String {
```
glbrntt, line 20: Should this be a computed property?

Sources/GRPCNIOTransportCore/Internal/Channel+AddressInfo.swift
```diff
@@ -0,0 +1,73 @@
...
+        return "unix:\(local.pathname!)"
+
+      case .v4, .v6:
+        // Remote address is UDS but local isn't. This shouldn't ever happen.
+        return "unix:<unknown>"
+      }
+    }
+  }
+
+  func getLocalAddressInfo() -> String {
```
glbrntt, line 54: Should this be a computed property?

Sources/GRPCNIOTransportCore/Client/Connection/Connection.swift
```diff
@@ -421,6 +421,14 @@ extension Connection {
...
+      self.http2Stream.channel.getRemoteAddressInfo()
```
glbrntt, line 425: I'm not sure whether the remote info is actually set on the HTTP/2 stream channel. 

Also given this information won't change during the lifetime of a single connection, we can get both of these values up front when we create the HTTP/2 connection. That way we can avoid recreating the strings for each stream (which is potentially expensive).

glbrntt (changes requested):
We should really add tests for this too. There are some already for the server checking its peer info: https://github.com/grpc/grpc-swift-nio-transport/blob/92ad3c0d60dbebec75368dc71790fd5db4100560/Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportTests.swift#L1630-L1658

The control service has a peer info RPC which just returns the peer info to the client as a string. We could change that type to contain the client and server local and remote peer info. The server can obviously extract that info from its own context and the client could inject it into metadata from an interceptor so that the server can then pull it out and stuff it back in the response.

Take a look here too: https://github.com/grpc/grpc-swift-nio-transport/blob/92ad3c0d60dbebec75368dc71790fd5db4100560/Tests/GRPCNIOTransportHTTP2Tests/ControlService.swift#L65-L75

Sources/GRPCNIOTransportCore/Client/Connection/Connection.swift
glbrntt: Yeah that works, I don't mind about the suffix so much. I'd like the consistency between "remote" and "local" (vs  "peer" and "local")

glbrntt (approved):
Looks good, assuming CI turns green :)

glbrntt (changes requested):

glbrntt (approved):

PR #57, opened 2025-01-16
Group docs and fix visibility

glbrntt:
Motivation:

Grouping docs in the DocC index is a useful way to make the docs more approachable.

Modifications:

- Group the docs
- Add a few more docs
- Change the visibility of an accidentally public type

Result:

Better docs


glbrntt:
This is technically an API break because a public type was made `package` _but_ there were no public inits so the chance of this actually breaking is close to zero.

gjcairo (approved):

PR #60, opened 2025-01-17
Fix deprecation warnings

glbrntt:


glbrntt:
This got included in https://github.com/grpc/grpc-swift-nio-transport/pull/55

PR #61, opened 2025-01-17
Move test JSON (de)serializer to its own file

gjcairo:
This is a nit, but tests across different files are using `JSONSerializer` and `JSONDeserializer`, but it was defined at the bottom of one set of tests.
For tidiness and to improve discoverability a bit, they should be in their own file.

### DIFF
--- a/Tests/GRPCNIOTransportHTTP2Tests/Test Utilities/JSONSerializing.swift
+++ b/Tests/GRPCNIOTransportHTTP2Tests/Test Utilities/JSONSerializing.swift
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2025, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


Tests/GRPCNIOTransportHTTP2Tests/Test Utilities/JSONSerializing.swift
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+import GRPCCore
+
+struct JSONSerializer<Message: Encodable>: MessageSerializer {
```
glbrntt, line 20: I think this'll need updating as the API changed (and was fixed in #55)

glbrntt (approved):

PR #73, opened 2025-02-18
Uncomment `StreamPriorityData` frames in stream handler tests

gjcairo:
We had TODOs to uncomment some H2 frames in tests when their init became available, which it has. This PR uncomments them.

### DIFF
--- a/Tests/GRPCNIOTransportCoreTests/Client/GRPCClientStreamHandlerTests.swift
+++ b/Tests/GRPCNIOTransportCoreTests/Client/GRPCClientStreamHandlerTests.swift
@@ -40,10 +40,9 @@ final class GRPCClientStreamHandlerTests: XCTestCase {
     let framesToBeIgnored: [HTTP2Frame.FramePayload] = [
       .ping(.init(), ack: false),
       .goAway(lastStreamID: .rootStream, errorCode: .cancel, opaqueData: nil),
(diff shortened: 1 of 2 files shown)


Tests/GRPCNIOTransportCoreTests/Client/GRPCClientStreamHandlerTests.swift
```diff
@@ -40,10 +40,9 @@ final class GRPCClientStreamHandlerTests: XCTestCase {
     let framesToBeIgnored: [HTTP2Frame.FramePayload] = [
       .ping(.init(), ack: false),
       .goAway(lastStreamID: .rootStream, errorCode: .cancel, opaqueData: nil),
-      // TODO: uncomment when it's possible to build a `StreamPriorityData`.
-      // .priority(
-      //   HTTP2Frame.StreamPriorityData(exclusive: false, dependency: .rootStream, weight: 4)
-      // ),
+      .priority(
+        HTTP2Frame.StreamPriorityData(exclusive: false, dependency: .rootStream, weight: 4)
```
glbrntt, line 44: Do we need to bump the minimum required http2 version as well?

gjcairo: No, we're already on the latest version

glbrntt (approved):

PR #76, opened 2025-02-26
Ensure port isn't included in SNI hostname

### LINKED ISSUE #71: Port can be incorrectly included in TLS SNI hostname
### Describe the bug

I am transitioning my iOS app to the new gRPC implementation.  However, I get the following error when trying to establish SSL with a private root CA:

~~~
The server accepted the TCP connection but closed the connection before completing the HTTP/2 connection preface." (cause: "handshakeFailed(NIOSSL.BoringSSLError.sslError([Error: 268435581 error:1000007d:SSL routines:OPENSSL_internal:CERTIFICATE_VERIFY_FAILED
~~~

The following code contains the configuration:
~~~
let caPath = Bundle.main.path(forResource: "root_ca.crt", ofType: nil)!
let tlsConfig = HTTP2ClientTransport.Posix.TransportSecurity.TLS(
    certificateChain: [],
    privateKey: nil,
    serverCertificateVerification: .fullVerification,
    trustRoots: .certificates( [.file(path: caPath, format: .pem)] )
)
~~~

### To reproduce

Configure grpc client with the TLS config above.

### Expected behaviour

Successful SSL connection

### Additional information

Here is the code before under gRPC 1.24 which was working:

~~~
var trustRoots = [NIOSSLCertificate]()
trustRoots.append(try! NIOSSLCertificate(file: Bundle.main.path(forResource: "root_ca.crt", ofType: nil)!, format: .pem))
let keepalive = ClientConnectionKeepalive(
    interval: .seconds(5),
    timeout: .seconds(2))
let channel = ClientConnection
    .usingTLSBackedByNIOSSL(on: group)
    .withTLS(trustRoots: .certificates(trustRoots))
    .withTLS(certificateVerification: .fullVerification)
    .withBackgroundActivityLogger(grpcLogger)
    .withKeepalive(keepalive)
    .connect(host: "vultr0.chrishong.dev", port: 50050)
~~~

glbrntt:
Motivation:

gRPC derives the authority from various sources and uses this value for the SNI server hostname. The authority should include non-standard ports and the SNI hostname must not include the port. In some cases this wasn't being respected and the port was being used in SNI this results in handshake failures.

Modifications:

- Have the Connection sanitize the authority so that it's suitable for SNI.

Result:

- Fewer handshake failures
- Resolves #71

### DIFF
--- a/Tests/GRPCNIOTransportCoreTests/Client/Connection/ConnectionTests.swift
+++ b/Tests/GRPCNIOTransportCoreTests/Client/Connection/ConnectionTests.swift
@@ -232,6 +232,11 @@ final class ConnectionTests: XCTestCase {
       authority: "foo.example-31415",
       expected: "foo.example-31415"
     )
+
+    try await self.testAuthorityIsSanitized(
+      authority: "foo.example.com:abc123",
+      expected: "foo.example.com:abc123"
+    )
   }
 }
 
--- a/Tests/GRPCNIOTransportCoreTests/Client/Connection/ConnectionTests.swift
+++ b/Tests/GRPCNIOTransportCoreTests/Client/Connection/ConnectionTests.swift
@@ -21,6 +21,7 @@ import NIOCore
 import NIOHPACK
 import NIOHTTP2
 import NIOPosix
+import Synchronization
 import XCTest
 
 final class ConnectionTests: XCTestCase {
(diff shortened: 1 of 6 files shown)


Tests/GRPCNIOTransportCoreTests/Client/Connection/ConnectionTests.swift
```diff
@@ -199,6 +200,39 @@ final class ConnectionTests: XCTestCase {
...
+      expected: "foo.example.com"
+    )
+
+    try await self.testAuthorityIsSanitized(
+      authority: "foo.example.com:31415",
+      expected: "foo.example.com"
+    )
+
+    try await self.testAuthorityIsSanitized(
+      authority: "foo.example-31415",
```
gjcairo, line 232: Not sure if this would be a valid authority name (not sure if there are rules around the SNI format), but could we test that something with a colon like `foo.example:a123` would not strip what comes after the colon?

gjcairo (approved):

PR #81, opened 2025-03-17
Update min version of grpc-swift to fix static Linux SDK build

gjcairo:
https://github.com/grpc/grpc-swift/releases/tag/2.1.1 fixes an issue that made `grpc-swift` not compile using the static Linux SDK. This caused issues when trying to build `grpc-swift-nio-transport` using the static Linux SDK. This PR updates the minimum `grpc-swift` version to make it build successfully.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -35,7 +35,7 @@ let products: [Product] = [
 let dependencies: [Package.Dependency] = [
   .package(
     url: "https://github.com/grpc/grpc-swift.git",
-    from: "2.0.0"
+    from: "2.1.1"
   ),
   .package(
     url: "https://github.com/apple/swift-nio.git",


glbrntt (commented):
I don't think we actually need to make this change; there's no new API we're depending on. SPM will pick up the new patch release of grpc-swift.

gjcairo:
Yes I wasn't sure this would be needed. But is it possible SPM resolves to the version without the patch when you're trying to build with the static SDK? I thought updating the version here would remove that possibility altogether.

glbrntt:
It's possible if there's an additional constraint in the dependency graph to limit the version of grpc-swift but that would also be problematic if we bumped the version here.

PR #85, opened 2025-04-07
Break timer retain cycle

### LINKED ISSUE #84: Server transport leaking
We're seeing continuously increasing memory usage in a production app with the Swift NIO transport.

It's quite obvious when looking at the memory graph while repeatedly calling an RPC:

Image

The Xcode memory graph debugger indicates leaks of `GRPCNIOTransportCore.Timer<GRPCNIOTransportCore.ServerConnectionManagementHandler.KeepaliveTimeoutHandlerView>` and `GRPCNIOTransportCore.Timer<GRPCNIOTransportCore.ServerConnectionManagementHandler.KeepaliveTimerHandlerView>`, as well as `ServerConnectionManagementHandler` and `Swift._SetStorage<NIOHTTP2.HTTP2StreamID>`.

Image

### Reproducing the issue

I have created a project to easily reproduce this

```bash
git clone https://github.com/Obbut/GRPCNIOTransportIssue.git
cd GRPCNIOTransportIssue
swift run Server
```

I added a script that repeatedly calls an RPC in the example project. To run it, you need [grpcurl](https://github.com/fullstorydev/grpcurl).

```bash
Scripts/Call-Server.sh
```

### Memory graph debugger

Just as a note, the memory graph debugger in Xcode (and profiling in Instruments) doesn't work out of the box for me. To fix this you can manually codesign the executable with the `com.apple.security.get-task-allow` entitlement.

glbrntt:
Motivation:

The timers used by the client/server connection management hanlders indirectly hold references to the underlying handler. The hanlder holds a reference to the timer. This cycle isn't broken so a handler is leaked for each connection.

Modifications:

- Nil out the timers to break the retain cycle when the channel becomes inactive

Result:

- Fewer leaks
- Resolves #84

### DIFF
--- a/Sources/GRPCNIOTransportCore/Client/Connection/ClientConnectionHandler.swift
+++ b/Sources/GRPCNIOTransportCore/Client/Connection/ClientConnectionHandler.swift
@@ -160,7 +160,14 @@ package final class ClientConnectionHandler: ChannelInboundHandler, ChannelOutbo
     }
 
     self.keepaliveTimerHandler?.cancel()
+    self.keepaliveTimerHandler = nil
+
     self.keepaliveTimeoutHandler?.cancel()
+    self.keepaliveTimeoutHandler = nil
+
+    self.maxIdleTimerHandler?.cancel()
+    self.maxIdleTimerHandler = nil
+
     context.fireChannelInactive()
   }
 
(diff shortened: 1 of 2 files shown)


Joannis:
Thanks for the quick resolution with this PR!

gjcairo (approved):

Issue #83, opened 2025-04-07
Assertion failure in GRPCStreamStateMachine when connecting to non-GRPC HTTP server


Obbut:
When connecting to a non-GRPC HTTP server (for example: `example.com:443`), an assertion failure is triggered:

```
GRPCNIOTransportCore/GRPCStreamStateMachine.swift:1092: Fatal error: Server cannot have sent a message before sending the initial metadata.

💣 Program crashed: System trap at 0x00000001988cf1f8
```

I have created a [reproduction project](https://github.com/Obbut/GRPCNIOTransportIssue). To reproduce the issue:

```bash
git clone https://github.com/Obbut/GRPCNIOTransportIssue.git
cd GRPCNIOTransportIssue
swift run Client
```

---

This scenario represents a misconfiguration or unexpected server response, which, although incorrect, probably shouldn't result in a crash/assertion failure.

glbrntt:
Just to note: this is only an assertion failure for debug builds. In release builds it'll be an error which will be handled gracefully.

Obbut:
> Just to note: this is only an assertion failure for debug builds. In release builds it'll be an error which will be handled gracefully.

I understand. During the development of a project for which the gRPC client connection is not essential, we encountered this issue. While it’s not a significant problem, the assertion failure disrupts the development workflow more than I’d prefer.

glbrntt:
Oh absolutely, this is a bug that needs fixing.

I just wanted to make it clear that this isn't a runtime crash in release mode which could be vulnerability.

gjcairo:
Hey @Obbut, little update here. As George pointed out, we are already throwing in release mode vs runtime crashing in debug mode. We think this pattern is generally the right one for unrecoverable states. However, for unexpected-but-not-unrecoverable states, such as getting an unexpected/malformed response from a server, we should just drop these on the floor and move on if we can.

#88 addresses a few bad state transitions where we were not ignoring some packages received from the server that we _should_ have ignored, because we had transitioned the server to a closed state anyways. For your particular case of connecting to a non-gRPC server, this patch should avoid the crash so I'll close this issue. If it doesn't, please feel free to reopen it. Thanks again for reporting!

PR #88, opened 2025-04-10
Fix bad transition when server sends invalid headers in stream state machine

gjcairo:
While triaging https://github.com/grpc/grpc-swift-nio-transport/issues/83 we discovered an issue with the stream state machine, where we would not transition the server to `closed` when receiving invalid headers. This is not correct, because at this point the server is in an unrecoverable state from the client's perspective, and we should ignore any subsequent messages.

This PR fixes this transition, which was causing an "invalid state" error (or runtime crash in debug mode). It now drops any further messages from the server on the floor, instead of throwing/sending another error message back to the client.

I've also fixed a problem with receiving headers with a 1xx status code: they should be ignored, but we were transitioning to closed states instead.

### DIFF
--- a/Sources/GRPCNIOTransportCore/GRPCStreamStateMachine.swift
+++ b/Sources/GRPCNIOTransportCore/GRPCStreamStateMachine.swift
@@ -831,6 +831,7 @@ extension GRPCStreamStateMachine {
   private enum ServerHeadersValidationResult {
     case valid
     case invalid(OnMetadataReceived)
+    case skip
   }
 
   private mutating func clientValidateHeadersReceivedFromServer(
@@ -863,7 +864,7 @@ extension GRPCStreamStateMachine {
         // For 1xx status codes, the entire header should be skipped and a
         // subsequent header should be read.
         // See https://github.com/grpc/grpc/blob/7f664c69b2a636386fbf95c16bc78c559734ce0f/doc/http-grpc-status-mapping.md
-        return .invalid(.doNothing)
+        return .skip
       }
 
       // Forward the mapped status code.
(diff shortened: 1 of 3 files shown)


Sources/GRPCNIOTransportCore/GRPCStreamStateMachine.swift
```diff
@@ -1153,9 +1161,13 @@ extension GRPCStreamStateMachine {
...
     case .clientOpenServerClosed, .clientClosedServerClosed:
-      try self.invalidState(
-        "Cannot have received anything from a closed server."
-      )
+      // If the server is closed, it's because it actually closed, or because the client
+      // transitioned it to a close state because it returned invalid headers.
+      // In either case, we will have already surfaced a status + trailers response to the client,
+      // so if we receive further packages, we should just drop them on the floor,
+      // as there's nothing for us to do with them.
+      return .doNothing
```
glbrntt, line 1169: There are presumably other places we need to make this change too, like receiving other events from the server? Can you audit for these?

Tests/GRPCNIOTransportCoreTests/Client/GRPCClientStreamHandlerTests.swift
```diff
@@ -161,6 +179,16 @@ final class GRPCClientStreamHandlerTests: XCTestCase {
...
       )
     )
+
+    // We should not throw if the server sends another message:
+    // we should drop it, since the server is now closed.
+    var buffer = ByteBuffer()
+    buffer.writeInteger(UInt8(0))  // not compressed
+    buffer.writeInteger(UInt32(42))  // message length
+    buffer.writeRepeatingByte(0, count: 42)  // message
+    let serverDataPayload = HTTP2Frame.FramePayload.Data(data: .byteBuffer(buffer), endStream: true)
```
glbrntt, line 189: What about the server sending trailers with end stream set? We should cover that case as well.

glbrntt (approved):
Thanks Gus!

PR #90, opened 2025-04-17
Add NIOTS bootstrap configurators to config

gjcairo:
`swift-nio-transport-services` added support for customising the `NWParameters` used in the underlying connections in https://github.com/apple/swift-nio-transport-services/pull/230.

This PR adds API in `grpc-swift-nio-transport` to allow users to customise the bootstraps used when creating gRPC connections using NIOTS. This will allow them to, for example, customise their `NWParameters` using the new APIs in `swift-nio-transport-services`.


gjcairo:
We've added this API to v1 in https://github.com/grpc/grpc-swift/pull/2223
We don't currently plan to add the same API to v2.

PR #92, opened 2025-04-23
Apply patch to clamp window size to Int32.max

jessezamora:
This prevents integer overflows from occurring on 32-bit architectures such as armv7. Could also work for i686, etc.

I also ran the test suite with this patch and all seems to work!


glbrntt (approved):
Awesome, thank you!

PR #97, opened 2025-05-09
Expose the peerCertificate from mTLS to ServerContext

jtdavey:
Motivation:

When using gRPC, it's useful to pull the peerCertificate from a successful mTLS connection for authn and/or authz.

Modifications:

These changes make use of the new transportSpecific field in the ServerContext. After a successful hanshake has completed, it uses the provided handler to retrieve the verified peer certificate and populate the Server Context.

There's also a simple happy path test that's been added for validation purposes.

Temporarily, this commit also references specific updates in swift-nio-ssl and grpc-swift.

Result:

This commit allows access to the swift Certificate object in an interceptor for the Posix HTTP2 transport. It also adds a mechanism for other transports to allow their own transport specific data.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -35,7 +35,7 @@ let products: [Product] = [
 let dependencies: [Package.Dependency] = [
   .package(
     url: "https://github.com/grpc/grpc-swift.git",
-    from: "2.0.0"
+    revision: "0d850d6"
   ),
   .package(
     url: "https://github.com/apple/swift-nio.git",
@@ -51,7 +51,7 @@ let dependencies: [Package.Dependency] = [
   ),
   .package(
     url: "https://github.com/apple/swift-nio-ssl.git",
-    from: "2.29.0"
+    revision: "874ad69"
   ),
   .package(
     url: "https://github.com/apple/swift-nio-extras.git",
@@ -60,6 +60,10 @@ let dependencies: [Package.Dependency] = [
   .package(
     url: "https://github.com/apple/swift-certificates.git",
     from: "1.5.0"
+  ),
+  .package(
+    url: "https://github.com/apple/swift-asn1.git",
+    from: "1.0.0"
   ),
 ]
 
@@ -111,6 +115,9 @@ let targets: [Target] = [
       .product(name: "GRPCCore", package: "grpc-swift"),
       .product(name: "NIOPosix", package: "swift-nio"),
       .product(name: "NIOSSL", package: "swift-nio-ssl"),
+      .product(name: "X509", package: "swift-certificates"),
+      .product(name: "SwiftASN1", package: "swift-asn1"),
+
     ],
     swiftSettings: defaultSwiftSettings
   ),
--- a/Package.swift
+++ b/Package.swift
@@ -117,7 +117,6 @@ let targets: [Target] = [
       .product(name: "NIOSSL", package: "swift-nio-ssl"),
       .product(name: "X509", package: "swift-certificates"),
       .product(name: "SwiftASN1", package: "swift-asn1"),
-
     ],
     swiftSettings: defaultSwiftSettings
   ),
--- a/Package.swift
+++ b/Package.swift
@@ -35,7 +35,7 @@ let products: [Product] = [
 let dependencies: [Package.Dependency] = [
   .package(
     url: "https://github.com/grpc/grpc-swift.git",
-    revision: "0d850d6"
+    from: "2.2.0"
   ),
   .package(
     url: "https://github.com/apple/swift-nio.git",
@@ -51,7 +51,7 @@ let dependencies: [Package.Dependency] = [
   ),
   .package(
     url: "https://github.com/apple/swift-nio-ssl.git",
-    revision: "874ad69"
+    from: "2.31.0"
   ),
   .package(
     url: "https://github.com/apple/swift-nio-extras.git",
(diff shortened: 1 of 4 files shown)


Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportTLSEnabledTests.swift
```diff
@@ -58,6 +59,63 @@ struct HTTP2TransportTLSEnabledTests {
...
+      var derSerializer = DER.Serializer()
+      try peerCertificate.serialize(into: &derSerializer)
+      #expect(derSerializer.serializedBytes == self.clientCert)
+      return try await next(request, context)
+    }
+  }
+
+  @Test(
+    "Using the mTLS defaults, and with Posix transport, validate we get the peer cert on the server",
+    arguments: TransportKind.supported
```
glbrntt, line 91:
```suggestion
    arguments: [TransportKind.posix]
```

Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportTLSEnabledTests.swift
```diff
@@ -58,6 +59,63 @@ struct HTTP2TransportTLSEnabledTests {
...
+  )
+  func testRPC_mTLS_TransportContext_OK(supportedTransport: TransportKind) async throws {
+    if TransportKind.posix == supportedTransport {
```
glbrntt, line 94: We can remove this if we change `arguments`

Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportTLSEnabledTests.swift
```diff
@@ -58,6 +59,63 @@ struct HTTP2TransportTLSEnabledTests {
...
+    func intercept<Input, Output>(
+      request: GRPCCore.StreamingServerRequest<Input>,
+      context: GRPCCore.ServerContext,
+      next: @Sendable (GRPCCore.StreamingServerRequest<Input>, GRPCCore.ServerContext) async throws
+        -> GRPCCore.StreamingServerResponse<Output>
+    ) async throws -> GRPCCore.StreamingServerResponse<Output>
+    where Input: Sendable, Output: Sendable {
+      let transportSpecific = context.transportSpecific
+      #expect(transportSpecific != nil)
+      #expect(transportSpecific is HTTP2ServerTransport.Posix.Context)
```
glbrntt, line 76: We don't need these as they're covered by the `#require` below

Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportTLSEnabledTests.swift
```diff
@@ -58,6 +59,63 @@ struct HTTP2TransportTLSEnabledTests {
...
+      let transportSpecificAsPosixContext = try #require(
+        transportSpecific as? HTTP2ServerTransport.Posix.Context
+      )
+      let peerCertificate = try #require(transportSpecificAsPosixContext.peerCertificate)
+      #expect(transportSpecificAsPosixContext.peerCertificate != nil)
```
glbrntt, line 81: Don't need this one: it's covered by the `#require` just above

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
```diff
@@ -186,6 +199,15 @@ extension HTTP2ServerTransport {
 }
 
 extension HTTP2ServerTransport.Posix {
+  /// Context for Posix TransportSpecific
+  public struct Context: ServerContext.TransportSpecific {
```
FranzBusch, line 203: How does this compose overall? Shouldn't this be something that really every transport can vend optionally so code that has to do authn/z doesn't have to become transport specific @glbrntt?

glbrntt: The longer term move is to have mixin protocols (e.g. `HasPeerCertificate`) which contexts would conform to. Any common code would then use these rather than relying on a concrete transport specific type.

FranzBusch: I guess my meta question is why do we consider that peer certificate needs to be a mix-in protocol and we can't make it part of the `ServerContext` right away. It seems in line with `remotePeer: String` to have a `remotePeerCertificate: Certificate?`. Since it has to be optional anyways.

glbrntt: It isn't really in line with it, you can't compute on `remotePeer` beyond using it as an identifier. `remotePeer` is also available to all transports, whereas the peer cert is only available here.

Adding it to the context brings in three more dependencies which would only be used by the NIO transport.

FranzBusch: Why only the NIO transport? Can't we provide a `Certificate` based peer identity on the `NIOTS` transport as well?

glbrntt: NIO transport includes NIOTS. The NIOTS transport work hasn't been done yet (I'm not sure if the right APIs currently exist to do it.)

Package.swift
```diff
@@ -35,7 +35,7 @@ let products: [Product] = [
 let dependencies: [Package.Dependency] = [
   .package(
     url: "https://github.com/grpc/grpc-swift.git",
-    from: "2.0.0"
+    revision: "0d850d6"
```
glbrntt, line 38: FYI https://github.com/grpc/grpc-swift/releases/tag/2.2.0

glbrntt (approved):

PR #101, opened 2025-05-15
Add a transport for wrapping a NIO Channel

glbrntt:
Motivation:

The NIO based client and transports want to fully manage the underlying NIO channels for you; creating them and tearing them down as necessary. However, in some instances users already have a NIO channel which they may also want to use a client. This might happen if they want to run gRPC within another protocol, for example.

Modifications:

- Add a "wrapped channel" client transport which takes ownership of a NIO channel and configures the channels pipeline to do gRPC.
- This is a different transport because a number of features aren't available such as reconnects and client-side load-balancing.

Result:

Users can wrap an existing NIO channel into a custom transport.


gjcairo (approved):
LGTM, just two nits.

PR #104, opened 2025-05-20
Add support for NIOCertificateReloading

glbrntt:
Motivation:

swift-nio-extras recently added support for certificate reloaders which can keep cert chain and private keys up-to-date in the background and provide current values when asked for them. We should expose an API allowing these to be configured.

Modifications:

- Expose API on the TLS configs for posix-based transports to create the TLS config using a cert reloader

Result:

Users can create TLS configs using the new cert relader

### DIFF
--- a/Sources/GRPCNIOTransportHTTP2Posix/Config+TLS.swift
+++ b/Sources/GRPCNIOTransportHTTP2Posix/Config+TLS.swift
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+private import GRPCCore
+public import NIOCertificateReloading
 public import NIOSSL
 
 extension HTTP2ServerTransport.Posix {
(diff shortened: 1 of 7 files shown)


Sources/GRPCNIOTransportHTTP2Posix/Config+TLS.swift
```diff
@@ -71,6 +106,39 @@ extension HTTP2ServerTransport.Posix {
       )
       return .tls(tlsConfig)
     }
+
+    /// Create a new TLS config suitable for mTLS using a certificate reloader to provide the
+    /// certificate chain and private key.
+    ///
+    /// The reloader must provide an initial certificate chain and private key. If you have already
```
rnro, line 113:
```suggestion
    /// The reloader must provide an initial certificate chain and private key. If you already
```

gjcairo (approved):

rnro (approved):

PR #114, opened 2025-06-04
fix dns resolver for android build grpc-swift 2

flymedllva:
Hi, I tried to build grpc-swift 2 for Android and ran into a problem that grpc-swift-nio-transport is not building.
Made a fix to fix the library, now grpc-swift-nio-transport and grpc-swift 2 builds successfully on Android.
Plus, the other day they added Wasm and Android support to the Swift Package Index: https://swiftpackageindex.com/blog/adding-wasm-and-android-compatibility-testing


glbrntt (approved):
Thank you!

PR #115, opened 2025-07-09
Load multiple certificates from a single trust root file

josephnoir:
### Motivation

When passing a PEM file with multiple certificates as trust roots, only the first one is loaded.

### Modifications

Instead of creating a NIOSSLCertificate directly, use a function that loads multiple certificates from a PEM file. Add a test case that creates a certificate chain in a temporary directory to verify the expected behavior.

### Result

Multiple certificates can be loaded from a single PEM file.

### DIFF
--- a/Tests/GRPCNIOTransportHTTP2Tests/Test Utilities/CertificateChain.swift
+++ b/Tests/GRPCNIOTransportHTTP2Tests/Test Utilities/CertificateChain.swift
@@ -19,17 +19,34 @@ import Foundation
 import SwiftASN1
 import X509
 
+/// Create a certificate chain with a root and intermediate certificate stored in a trust roots file and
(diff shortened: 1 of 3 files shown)


Tests/GRPCNIOTransportHTTP2Tests/Test Utilities/CertificateChain.swift
```diff
@@ -0,0 +1,316 @@
...
+    let key: Certificate.PrivateKey
+  }
+
+  /// Leaf certificates can authenticate either a client or a server
+  enum Authenticating {
+    case client
+    case server
+  }
+
+  /// Writing the files to disk returns a dictionary with these keys to access the file locations
```
glbrntt, line 49: Rather than a dictionary can we just have a `struct` with a property for each of these? We can avoid the `!` when accessing them each time then.

Issue #116, opened 2025-07-11
Better host resolving errors handling (Channel isn't ready.)


alex-vasenin-peggy:
Guys, I spent an hour figuring out the cause of the following error:
> unavailable: "Channel isn't ready."

Turns out, instead of this:
```
let transport = try! HTTP2ClientTransport.TransportServices(target: .dns(host: "grpc.mydomain.com"), transportSecurity: .tls)
```

I've been using this:
```
let transport = try! HTTP2ClientTransport.TransportServices(target: .ipv4(host: "grpc.mydomain.com"), transportSecurity: .tls)
```

If you expect only IPv4 address in `.ipv4(host:)` you should crash (or throw) if you get something else, and frankly `.ipv4(address:)` communicates the idea much better. Sorry for venting out.

alex-vasenin-peggy:
Thank you, that's very helpful! 👍

PR #117, opened 2025-07-14
Deprecate ambiguous IPv4 and IPv6 target APIs

glbrntt:
Motivation:

The API for creating IPv4/IPv6 resolvable targets is a bit ambiguous: it's quite easy to assume they will only resolve IPv4 or IPv6 addresses. Instead they expect to be provided with resolved IPv4 or IPv6 addresses.

Modifications:

- Deprecate these APIs and replace them with more obvious spellings
- Add a debug-only check to validate that the target is a valid IPv4/IPv6 address

Result:

- Harder to make mistakes
- Easier to diagnose mistakes

### DIFF
--- a/Sources/GRPCNIOTransportCore/Client/Resolver/NameResolver+IPv6.swift
+++ b/Sources/GRPCNIOTransportCore/Client/Resolver/NameResolver+IPv6.swift
@@ -67,7 +67,7 @@ extension ResolvableTarget where Self == ResolvableTargets.IPv6 {
 
   /// Creates a new resolvable IPv6 target for a single address.
   /// - Parameters:
-  ///   - host: The resolved host address.
+  ///   - address: The resolved host address.
   ///   - port: The port on the host.
   /// - Returns: A ``ResolvableTarget``.
   @available(gRPCSwiftNIOTransport 2.1, *)
--- a/Sources/GRPCNIOTransportCore/Client/Resolver/NameResolver+IPv6.swift
+++ b/Sources/GRPCNIOTransportCore/Client/Resolver/NameResolver+IPv6.swift
@@ -15,6 +15,7 @@
  */
 
 internal import GRPCCore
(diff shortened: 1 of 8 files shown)


Sources/GRPCNIOTransportCore/Client/Resolver/NameResolver+IPv6.swift
```diff
@@ -29,6 +30,23 @@ extension ResolvableTargets {
     /// Create a new IPv6 target.
     /// - Parameter addresses: The IPv6 addresses.
     public init(addresses: [SocketAddress.IPv6]) {
+      debugOnly {
```
gjcairo, line 33: Wouldn't it also be useful to get some better error other than [the current](https://github.com/grpc/grpc-swift-nio-transport/issues/116#issue-3223783756) `unavailable: "Channel isn't ready."` in release builds, too? 

I'm thinking if the address comes from some config and there are different values for debug and release, it's possible that debug builds won't fail but then you'd get some confusing error in release.

glbrntt: It would be useful but at that layer you don't have enough information to tie it back to creating the wrong resolver. The "Channel isn't ready." message can happen for any number of reasons, including you could've correctly used a resolver but the server just isn't up yet.

gjcairo: I guess my question was actually if we should _always_ perform this check, not just in debug mode (throwing an error instead of failing an assertion), or if it's too expensive to afford it in this codepath.

glbrntt: I don't think it's necessary tbh, the change in spelling is clear enough.

gjcairo (approved):

PR #118, opened 2025-07-14
Fixed explicit Sendable issues.

rafaelcepeda:
Performed the necessary changes to enable compiler flag -require-explicit-sendable.


glbrntt (approved):
Thank you!

PR #122, opened 2025-08-03
Fix hanging graceful channel shutdown by cleaning up state.

k3zi:
Corrects two scenarios where stale state in the load balancer could prevent a graceful channel shutdown.

Previously, multiple endpoint updates or changes to the load balancer could leave undiscarded subchannels or old load balancers in the channel's state, blocking a clean shutdown.

This change ensures that previous subchannels and load balancers are fully removed from state when they shutdown. This also corrects a related test, which had been passing incorrectly due to an explicit cancellation masking the issue.


glbrntt (approved):
Thank you!

PR #134, opened 2025-10-16
Allow using a customVerificationCallback for clients verifying servers

Joannis:
We have a use case where a gRPC server runs locally with a custom CA. I want to verify some attributes in that certificate prior to connecting.


Joannis:
Thanks for the heads up @Lukasa

PR #175, opened 2026-04-16
Add flush coalescing

glbrntt:
Motivation:

As RPCs are backed by NIOAsyncChannel a lot of the natural structure for coalescing flush events is lost: each write from the NIOAsyncChannel results in a flush and as that has to execute onto the event loop it typically results in one flush per-write. Flushes can be coalesced if multiple happen on the same loop tick but that's very rare.

Modifications:

- Add a coalescing handler which coalesces flushes in the http/2 connection channel by delaying a flush by a maximum deadline or until enough bytes are written. This can dramatically decrease the number of syscalls.
- Plumb this through to be configurable

Result:

- Fewer flushes, better perf for many workloads

### DIFF
--- a/Sources/GRPCNIOTransportCore/Internal/FlushCoalescingHandler.swift
+++ b/Sources/GRPCNIOTransportCore/Internal/FlushCoalescingHandler.swift
@@ -1,0 +1,159 @@
+/*
+ * Copyright 2026, gRPC Authors All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import NIOCore
+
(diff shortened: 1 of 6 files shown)


Sources/GRPCNIOTransportCore/Internal/FlushCoalescingHandler.swift
```diff
@@ -0,0 +1,159 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import NIOCore
+
+internal final class FlushCoalescingHandler: ChannelDuplexHandler {
```
gjcairo, line 19: I wonder if this could be beneficial outside of gRPC? Do you think we could have this handler as public API in NIO perhaps? 

Or have you considered even adding this functionality to the `NIOAsyncChannel` itself perhaps? As it's trying to solve an issue with how it works, really. We could add this handler when wrapping a channel in a `NIOAsyncChannel`.

gjcairo (commented):
LGTM overall but left a question

Sources/GRPCNIOTransportCore/Internal/FlushCoalescingHandler.swift
glbrntt: Yeah, it's potentially helpful in some other applications. I think we need some experience with it before making it generally available though. 

It works well here because we're coalescing writes across multiple streams (the coalescing handler is in the connection channel) but if we added it to the stream channels instead I suspect the performance would be worse in many cases.

gjcairo (approved):

PR #177, opened 2026-05-14
 Add `.customSecure` transport security option to `HTTP2ClientTransport`.

dmaclach:
Introduces a new `.customSecure` option to `HTTP2ClientTransport.Posix.TransportSecurity`
and `HTTP2ClientTransport.TransportServices.TransportSecurity`, providing support for
custom secure transports without duplicating channel initialization logic.

It ensures:
- The transport layer treats the connection as secure (setting the pipeline scheme to
  .https appropriately).
- Standard TLS setup (such as NIOSSLHandler in Posix or tlsOptions in TransportServices)
  is skipped.
- Users can inject their custom security handlers using the existing standard
  `Config.channelDebuggingCallbacks.onCreateTCPConnection` debug callback
  (e.g., by adding handlers at position: .first).


glbrntt (changes requested):
Thanks for the PR! Adding a custom channel initializer with the TLS options is duplicative of the existing [channel debugging callback API](https://github.com/grpc/grpc-swift-nio-transport/blob/8412ddd18fd6dc25cbcbdaf6e288412115555194/Sources/GRPCNIOTransportCore/Client/HTTP2ClientTransport.swift#L204). I think we can support your use case by using that in conjunction with a smaller change to the `TLS` config.

For plaintext connections you wouldn't need to make any modifications to the existing API. 

For TLS I believe you just need to propagate the intention that something custom is being done so that the scheme is set appropriately and the `NIOSSLHandler` isn't added / `tlsOptions` isn't set. I think that boils down to having a `.customTLS` case (or similar) rather than `.custom(isPlaintext:_:)`. It'll need some docs explaining what happens when it's set and how users can use it appropriately though (i.e. in conjunction with the debugging API) as it's quite an easy API to hold incorrectly as errors won't surface until runtime.

Generally speaking modifying the channel pipeline is something we try to push users away from as the channel handlers are an implementation detail, but it seems like your use case relies on this. That's fine, but just beware that your channel pipeline modification code needs to be robust, especially if it's trying to add handlers in a position relative to an existing hander.

dmaclach:
Thanks @glbrntt ! Apologies.. the first version that I pushed was based on a much older version of `grpc-swift-nio-transport` that we had in an internal repo.  We were unaware that it had not been updating as we expected. When I did the merge I didn't pay enough attention to the diffs to realize that I should have changed it. I'm hoping this version is more to your liking.

Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportNIOPosixTests.swift
```diff
@@ -665,4 +665,70 @@ private struct LoopbackListenerFactory: HTTP2ServerTransport.ListenerFactory {
...
+          return self._called
+        }
+        set {
+          self.lock.lock()
+          defer { self.lock.unlock() }
+          self._called = newValue
+        }
+      }
+    }
+    let tracker = Tracker()
```
glbrntt: nit: if you `import Synchronization` you can just use `Mutex<Bool>`

dmaclach:
Broken formatting fixed. Do we need a new API version  for the case?

Sources/GRPCNIOTransportHTTP2TransportServices/Config+TLS.swift
```diff
@@ -174,10 +175,22 @@ extension HTTP2ClientTransport.TransportServices {
...
     }
 
     package let wrapped: Wrapped
 
+    /// Secure the connection with a custom secure transport mechanism.
+    ///
+    /// When using this case, you should also configure
+    /// `Config.channelDebuggingCallbacks.onCreateTCPConnection` to inject your custom security
+    /// handlers into the pipeline.
+    public static let customSecure = Self(wrapped: .customSecure)
```
glbrntt, line 194:
```suggestion
    @available(gRPCSwiftNIOTransport 2.8, *)
    public static let customSecure = Self(wrapped: .customSecure)
```

Sources/GRPCNIOTransportHTTP2Posix/Config+TLS.swift
```diff
@@ -274,6 +279,13 @@ extension HTTP2ClientTransport.Posix {
       Self(wrapped: .tls(tls))
     }
 
+    /// Secure the connection with a custom secure transport mechanism.
+    ///
+    /// When using this case, you should also configure
+    /// `Config.channelDebuggingCallbacks.onCreateTCPConnection` to inject your custom security
+    /// handlers into the pipeline.
+    public static let customSecure = Self(wrapped: .customSecure)
```
glbrntt, line 289:
```suggestion
    @available(gRPCSwiftNIOTransport 2.8, *)
    public static let customSecure = Self(wrapped: .customSecure)
```

Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportNIOPosixTests.swift
```diff
@@ -665,4 +665,70 @@ private struct LoopbackListenerFactory: HTTP2ServerTransport.ListenerFactory {
         connectionConfigurator.configure(channel: channel, tls: .none)
       }
   }
+  func testCustomTransportSecurityCallbackInvoked() async throws {
```
glbrntt: I don't think this tests validate anything different to the existing ChannelDebugCallbackTests. I think the only difference between `customSecure` and `plaintext` (assuming no debug callbacks are set) is that `customSecure` will set `:scheme` to `https`.

Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportNIOPosixTests.swift
```diff
@@ -665,4 +665,70 @@ private struct LoopbackListenerFactory: HTTP2ServerTransport.ListenerFactory {
...
+        set {
+          self.lock.lock()
+          defer { self.lock.unlock() }
+          self._called = newValue
+        }
+      }
+    }
+    let tracker = Tracker()
+
+    try await withThrowingDiscardingTaskGroup { group in
```
glbrntt: FWIW: these tests can be simplified a lot using the `withGRPC{Client,Server}` helpers, something like:

```swift
try await withGRPCServer(transport: ..., services: [...]) { server in 
  let address = try await server.listeningAddress
  try await withGRPCClient(transport: ...) { client in 
    // ...
  }
}
```

They'll shutdown the client/server when their respective closures return which is quite convenient for tests. I suspect most tests in this file were added before those helpers.

glbrntt:
> Broken formatting fixed. Do we need a new API version for the case?

Yes please for new public API

dmaclach:
> > Broken formatting fixed. Do we need a new API version for the case?
> 
> Yes please for new public API

OK.. hoping I did that right.

Sources/GRPCNIOTransportHTTP2Posix/Config+TLS.swift
```diff
@@ -274,6 +280,14 @@ extension HTTP2ClientTransport.Posix {
       Self(wrapped: .tls(tls))
     }
 
+    /// Secure the connection with a custom secure transport mechanism.
+    ///
+    /// When using this case, you should also configure
+    /// `Config.channelDebuggingCallbacks.onCreateTCPConnection` to inject your custom security
+    /// handlers into the pipeline.
+    @available(gRPCSwiftNIOTransport 2.9, *)
```
glbrntt:
```suggestion
    @available(gRPCSwiftNIOTransport 2.8, *)
```

Sources/GRPCNIOTransportHTTP2Posix/Config+TLS.swift
```diff
@@ -262,6 +262,12 @@ extension HTTP2ClientTransport.Posix {
     package enum Wrapped: Sendable {
       case plaintext
       case tls(TLS)
+      /// Custom secure transport.
+      ///
+      /// Use this in conjunction with `Config.channelDebuggingCallbacks.onCreateTCPConnection`
+      /// to inject custom channel handlers for security.
+      @available(gRPCSwiftNIOTransport 2.9, *)
```
glbrntt:
```suggestion
      @available(gRPCSwiftNIOTransport 2.8, *)
```

Sources/GRPCNIOTransportHTTP2TransportServices/Config+TLS.swift
```diff
@@ -174,10 +175,24 @@ extension HTTP2ClientTransport.TransportServices {
     package enum Wrapped: Sendable {
       case plaintext
       case tls(TLS)
+      /// Custom secure transport.
+      ///
+      /// Use this in conjunction with `Config.channelDebuggingCallbacks.onCreateTCPConnection`
+      /// to inject custom channel handlers for security.
+      @available(gRPCSwiftNIOTransport 2.9, *)
```
glbrntt:
```suggestion
      @available(gRPCSwiftNIOTransport 2.8, *)
```

Sources/GRPCNIOTransportHTTP2TransportServices/Config+TLS.swift
```diff
@@ -174,10 +175,24 @@ extension HTTP2ClientTransport.TransportServices {
...
     }
 
     package let wrapped: Wrapped
 
+    /// Secure the connection with a custom secure transport mechanism.
+    ///
+    /// When using this case, you should also configure
+    /// `Config.channelDebuggingCallbacks.onCreateTCPConnection` to inject your custom security
+    /// handlers into the pipeline.
+    @available(gRPCSwiftNIOTransport 2.9, *)
```
glbrntt:
```suggestion
    @available(gRPCSwiftNIOTransport 2.8, *)
```

glbrntt:
> > > Broken formatting fixed. Do we need a new API version for the case?
> > 
> > 
> > Yes please for new public API
> 
> OK.. hoping I did that right.

2.8 hasn't been released yet so I changed fixed up your annotations from 2.9 to 2.8.

There are a few comments still outstanding in the tests but otherwise this looks good.

dmaclach:
> > > > Broken formatting fixed. Do we need a new API version for the case?
> > > 
> > > 
> > > Yes please for new public API
> > 
> > 
> > OK.. hoping I did that right.
> 
> 2.8 hasn't been released yet so I changed fixed up your annotations from 2.9 to 2.8.

Thank you

Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportNIOPosixTests.swift
dmaclach: Ack. Removed the tests.

glbrntt (approved):
Thanks @dmaclach!

dmaclach:
@glbrntt Sorry for being dumb, but I'm not quite sure what to do about https://github.com/grpc/grpc-swift-nio-transport/actions/runs/26109607361/job/76786282484? Am I missing something obvious?

glbrntt:
> @glbrntt Sorry for being dumb, but I'm not quite sure what to do about https://github.com/grpc/grpc-swift-nio-transport/actions/runs/26109607361/job/76786282484? Am I missing something obvious?

It's a false positive; nothing for you to do. I'll merge over it.

dmaclach:
> > @glbrntt Sorry for being dumb, but I'm not quite sure what to do about https://github.com/grpc/grpc-swift-nio-transport/actions/runs/26109607361/job/76786282484? Am I missing something obvious?
> 
> It's a false positive; nothing for you to do. I'll merge over it.

🙏🏻

PR #178, opened 2026-05-20
Add ability to provide custom connector to HTTP2ClientTransport.

dmaclach:
This requires making HTTP2Connector and HTTP2Connection public.


dmaclach:
Hurrah.. realized that I could use the channelDebuggingCallbacks to do what I need and we can drop these pull requests.

PR #179, opened 2026-05-20
Change access modifiers

dmaclach:
Motivation: to allow clients to implement custom `ChannelHandler` and `HTTP2Connector` implementations.


glbrntt:
> Motivation: to allow clients to implement custom `ChannelHandler` and `HTTP2Connector` implementations.

Can you elaborate on specifically what you want to do in your custom connector and handler? This isn't API we want to make public, so perhaps there's an alternative way we can make your goals achievable.

dmaclach:
Hurrah.. realized that I could use the channelDebuggingCallbacks to do what I need and we can drop these pull requests.

PR #181, opened 2026-06-05
Resume stream continuation if LB is ready

glbrntt:
Motivation:

When requesting a stream from GRPCChannel, if it isn't yet ready a continuation will be enqueued. When it becomes ready the continuation will be dequeued and resumed. However, there's a small timing window between the state machine telling the caller to join the queue and the caller creating the continuation and enqueing it. During this window the channel could become ready and the continuation would never be resumed.o

Modifications:

- Resume the continuation if the channel is in the ready state during the call to enqueue

Result:

Fewer hangs


Lukasa (approved):
Shiny, I love it.

PR #185, opened 2026-07-20
Honour certificate verification mode when custom trust roots are set (TransportServices)

### LINKED ISSUE #184: TransportServices client TLS: serverCertificateVerification is ignored when custom trustRoots are set
## Description

`HTTP2ClientTransport.TransportServices.TLS` exposes `serverCertificateVerification` (e.g. `.noHostnameVerification`), but when `trustRoots` is set to `.certificates(...)`, the setting has no effect on how the peer certificate is actually evaluated.

In `Config+TLS.swift`, `TLSConfig.TrustRootsSource.makeTrustRootsConfig()` installs a `sec_protocol_verify_t` block that anchors the custom certificates and then calls `SecTrustEvaluateAsyncWithError` on the `SecTrust` handed over by Network.framework — which carries the default SSL server policy, including hostname matching. `.noHostnameVerification` only sets `sec_protocol_options_set_peer_authentication_required(true)` in `HTTP2ClientTransport+TransportServices.swift`; the verify block never consults the verification mode.

## Impact

Any deployment using a private CA whose server certificates don't carry a dNSName/iPAddress SAN matching the dialed host (e.g. certs pinned by CA with identity carried in a URI SAN, or clients dialing raw IPs that aren't in the cert) fails the handshake on every connection, even though the identical configuration works with `HTTP2ClientTransport.Posix`, where `.noHostnameVerification` maps to NIOSSL's `certificateVerification = .noHostnameVerification` (chain validation without hostname checks).

Confirmed empirically with such a certificate: `SecTrustEvaluateWithError` under `SecPolicyCreateSSL(true, hostname)` fails with `-67602` ("SSL hostname does not match name(s) in certificate"), while `SecPolicyCreateSSL(true, nil)` and `SecPolicyCreateBasicX509()` both trust the same chain.

## Suggested fix

In the custom-trust-roots verify block, honour the configured verification mode — e.g. for `.noHostnameVerification`, replace the trust's policy before evaluating:

```swift
SecTrustSetPolicies(actualTrust, SecPolicyCreateSSL(true, nil))
```

keeping the current behaviour for `.fullVerification`.

A fix is proposed in #185.

dmonagle:
Fixes #184.

## Motivation

With `trustRoots: .certificates(...)`, the TransportServices transports install a verify block that anchors the custom roots and then calls `SecTrustEvaluateAsyncWithError` on the trust exactly as Network.framework configured it — with the default SSL policy, which includes hostname matching. The configured `CertificateVerification` mode is never consulted there:

- `.noHostnameVerification` behaves like full verification: any certificate whose SANs don't cover the dialled endpoint is rejected (`errSSLBadCert`), even though the chain validates against the custom roots. This breaks private-CA deployments whose certs deliberately carry no DNS/IP SANs, and peers dialled by IP address. The identical configuration works with the Posix transport, where the mode maps to NIOSSL's `certificateVerification = .noHostnameVerification`.
- `.noVerification` still installs the evaluating block and can fail the handshake despite verification being explicitly disabled.

## Modifications

The verify block now receives the verification mode and whether it is evaluating server or client certificates:

- `.noHostnameVerification`: the trust is re-policied with `SecPolicyCreateSSL(<role>, nil)` before evaluation — chain validation against the custom anchors is unchanged, hostname matching is skipped.
- `.noVerification`: no verify block is installed (the callers already set `sec_protocol_options_set_peer_authentication_required(false)` for this mode), matching NIOSSL semantics.
- `.fullVerification`: behaviour unchanged.

## Result

`.noHostnameVerification` + custom trust roots connects when the chain is valid but the hostname doesn't match, on both client and server TransportServices transports, matching the Posix transport.

New regression test (`testClientNoHostnameVerificationIgnoresHostname`) covers all client/server transport combinations; without the source change the two TransportServices-client cases fail with `NWError.tls(-9808)`. Existing TLS suites pass (`HTTP2TransportTLSEnabledTests`, `TLSConfigurationTests`).

### DIFF
--- a/Sources/GRPCNIOTransportHTTP2TransportServices/Config+TLS.swift
+++ b/Sources/GRPCNIOTransportHTTP2TransportServices/Config+TLS.swift
@@ -324,8 +324,22 @@ extension HTTP2ClientTransport.TransportServices {
 
 @available(gRPCSwiftNIOTransport 2.0, *)
 extension NWProtocolTLS.Options {
-  func setUpVerifyBlock(trustRootsSource: TLSConfig.TrustRootsSource) {
-    if let (verifyQueue, verifyBlock) = trustRootsSource.makeTrustRootsConfig() {
+  func setUpVerifyBlock(
+    trustRootsSource: TLSConfig.TrustRootsSource,
+    verification: TLSConfig.CertificateVerification,
+    evaluatingServerCertificates: Bool
+  ) {
+    // With verification disabled there is nothing to evaluate; installing
+    // the block would evaluate the peer's chain anyway and could fail the
+    // handshake despite verification being explicitly turned off.
+    if verification == .noVerification {
+      return
+    }
+
+    if let (verifyQueue, verifyBlock) = trustRootsSource.makeTrustRootsConfig(
+      verification: verification,
+      evaluatingServerCertificates: evaluatingServerCertificates
+    ) {
       sec_protocol_options_set_verify_block(
         self.securityProtocolOptions,
         verifyBlock,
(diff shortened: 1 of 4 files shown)


glbrntt (approved):
Thank you!

dmonagle:
The two failing jobs (Linux nightly-main / nightly-next) reproduce on `main` — the nightly toolchain's SwiftPM now errors on the `Documentation.docc` directories ("found 1 file(s) which are unhandled; explicitly declare them as resources or exclude from the target") for four targets. Unrelated to this change; all release-toolchain jobs pass.

PR #186, opened 2026-07-20
Don't use IP literals for TLS SNI

dmonagle:
## Motivation

When a connection's authority contains an IP literal — e.g. a `.dns` target created with an IP address, which is common when callers resolve addresses themselves (multi-homed servers, NAT64 synthesis on mobile clients, health checks by address) — the literal is passed to the TLS handshake as the SNI server hostname. RFC 6066 § 3 only permits DNS hostnames in SNI, and NIOSSL enforces this: the connection fails with `NIOSSLExtraError.cannotUseIPAddressInSNI` before the handshake starts, surfacing to callers as `unavailable: Could not establish a connection…`.

`SocketAddress.sniHostname` already applies exactly this rule ("Literal IP addresses aren't allowed in the SNI hostname") — but the authority-derived hostname takes precedence over it and doesn't.

## Modifications

`Connection.sanitizeAuthorityForSNI` now returns `nil` for IP literals — bare IPv4/IPv6, with a trailing port, or bracketed IPv6 — after its existing port-stripping. The literal check runs before port-stripping too, since that logic would otherwise mangle bare IPv6 literals like `::1`. `sniServerHostname` then falls back through the existing `?? self.address.sniHostname` path, so such connections are made without SNI instead of failing.

## Result

Dialling an IP literal over TLS connects (subject to the configured certificate verification) instead of failing with `cannotUseIPAddressInSNI`. Hostname authorities are unaffected.

Extended the existing `testAuthorityIsSanitized` harness with IP-literal cases (`127.0.0.1`, `127.0.0.1:9443`, `::1`, `[2001:db8::1]:443`), all expecting no SNI.

### DIFF
--- a/Sources/GRPCNIOTransportCore/Client/Connection/Connection.swift
+++ b/Sources/GRPCNIOTransportCore/Client/Connection/Connection.swift
@@ -115,18 +115,39 @@ package final class Connection: Sendable {
     self.event.stream
   }
 
-  private static func sanitizeAuthorityForSNI(_ authority: String) -> String {
+  private static func sanitizeAuthorityForSNI(_ authority: String) -> String? {
+    // A bare IP literal is never a valid SNI hostname (RFC 6066 § 3 permits
+    // only DNS hostnames, and NIOSSL enforces this by failing the connection
+    // before the handshake starts). Check before stripping a port: the
+    // stripping below would otherwise mangle bare IPv6 literals like "::1".
+    if Self.isIPLiteral(authority) {
+      return nil
+    }
+
     // Strip off a trailing ":{PORT}". Look for the last non-digit byte, if it's
     // a colon then keep everything up to that index.
     let index = authority.utf8.lastIndex { byte in
       return byte < UInt8(ascii: "0") || byte > UInt8(ascii: "9")
     }
 
+    var host: String
     if let index = index, authority.utf8[index] == UInt8(ascii: ":") {
-      return String(authority.utf8[..<index])!
+      host = String(authority.utf8[..<index])!
     } else {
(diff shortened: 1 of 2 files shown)


Sources/GRPCNIOTransportCore/Client/Connection/Connection.swift
```diff
@@ -115,18 +115,39 @@ package final class Connection: Sendable {
...
+      host = String(authority.utf8[..<index])!
     } else {
-      return authority
+      host = authority
+    }
+
+    // Undo the bracketing applied to IPv6 addresses in authorities
+    // (e.g. "[2001:db8::1]:443").
+    if host.utf8.first == UInt8(ascii: "["), host.utf8.last == UInt8(ascii: "]") {
+      host = String(host.dropFirst().dropLast())
```
glbrntt: Square brackets imply an IPv6 address so we can just return `nil` here, I think.

glbrntt (changes requested):
One tiny change but looks good otherwise

Sources/GRPCNIOTransportCore/Client/Connection/Connection.swift
dmonagle: Done — brackets now return nil directly (and `host` became a `let`).

glbrntt (approved):
Great, thank you!

glbrntt:
Apologies for missing this: it looks like the CI is failing because of a Swift issue: https://github.com/swiftlang/swift/issues/85427

We might be able to workaround it with a patch to swift-nio: https://github.com/apple/swift-nio/pull/3696

PR #187, opened 2026-07-24
Cancel connection backoff on graceful shutdown

hamzahrmalik:
## Problem

When a gRPC client's connection is stuck in backoff (e.g. the server is unreachable), graceful shutdown would wait out that in-flight backoff sleep before returning — so shutdown could hang for up to the max backoff (120s by default), even with zero in-flight RPCs. In practice that means processes that are slow to exit on shutdown

## Cause

A subchannel that is sleeping in connection backoff has no established connection and no in-flight RPCs to drain. Its backoff `Task.sleep` was spawned into the subchannel's discarding task group, which graceful shutdown awaits rather than cancels, so `run()` (and therefore`GRPCClient.runConnections()`) did not return until the backoff elapsed — up to the default max of 120s — even with zero in-flight RPCs. This also contradicts `runConnections()`'s documented contract of returning once graceful shutdown is requested and in-flight RPCs have finished.

## Changes

Spawn the backoff via `addCancellableTask` and store its handle on the `.connecting` state (mirroring how the name-resolver task handle is tracked). When `shutDown()` is called while backing off, cancel that task and finish immediately; when a connect attempt is genuinely in-flight, behaviour is unchanged.

The new test fails without this fix, and passes after it.

### DIFF
--- a/Sources/GRPCNIOTransportCore/Client/Connection/LoadBalancers/Subchannel.swift
+++ b/Sources/GRPCNIOTransportCore/Client/Connection/LoadBalancers/Subchannel.swift
@@ -250,7 +250,9 @@ extension Subchannel {
       self.event.continuation.yield(.connectivityStateChanged(.shutdown))
       connection.close()
 
-    case .emitShutdownAndFinish:
+    case .emitShutdownAndFinish(let backoffHandle):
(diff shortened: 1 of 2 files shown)


glbrntt (approved):
Thanks @hamzahrmalik, nice patch.

PR #188, opened 2026-07-27
Update links and references to point to grpc-swift-2

samuelmurray:
There were a few outdated links from when `grpc/grpc-swift` was the repo in use.
Hopefully the search was thorough, used the following regexp: `"grpc-swift(?!-)"`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@ DerivedData/
 .swiftpm
 build
 /protoc-gen-swift
-/protoc-gen-grpc-swift
+/protoc-gen-grpc-swift-2
 third_party/**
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -6,7 +6,7 @@ DerivedData/
 .swiftpm
 build
 /protoc-gen-swift
-/protoc-gen-grpc-swift
+/protoc-gen-grpc-swift-2
```
glbrntt:
```suggestion
/protoc-gen-grpc-swift
```

glbrntt (approved):
thank you!

PR #189, opened 2026-07-28
Regenerate integration test sources from proto

samuelmurray:
Now that `grpc-swift-protobuf` is updated, the generated code differers slightly.

Other changes:
* Use vendored protoc in script
* Fetch updates to .proto from upstream
* Fix broken path to /Generated
* Add help script to check generated code

### DIFF
--- a/dev/check-generated-code.sh
+++ b/dev/check-generated-code.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+## Copyright 2024, gRPC Authors All rights reserved.
+##
+## Licensed under the Apache License, Version 2.0 (the "License");
+## you may not use this file except in compliance with the License.
+## You may obtain a copy of the License at
+##
+##     http://www.apache.org/licenses/LICENSE-2.0
+##
+## Unless required by applicable law or agreed to in writing, software
+## distributed under the License is distributed on an "AS IS" BASIS,
+## WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+## See the License for the specific language governing permissions and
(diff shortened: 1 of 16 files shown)


dev/check-generated-code.sh
```diff
@@ -0,0 +1,31 @@
...
+
+log() { printf -- "** %s\n" "$*" >&2; }
+error() { printf -- "** ERROR: %s\n" "$*" >&2; }
+fatal() { error "$@"; exit 1; }
+
+here="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# Re-generate everything.
+log "Regenerating protos..."
+"$here"/protos/generate.sh
```
samuelmurray, line 26: Perhaps this ought to first run `fetch.sh` to ensure that `upstream` is up-to-date?

samuelmurray:
Pretty sure the failing tests on `Linux (nightly main)` is unrelated to my changes. Let me know if I should address it here, or if we ignore it.
EDIT: Ah I see now that that check is not required, and that it failed in https://github.com/grpc/grpc-swift-nio-transport/pull/189 as well.

dev/check-generated-code.sh
samuelmurray: On the other hand, that might be too strict since it would fail builds (if this was part of PR checks) any time the upstream was updated..

glbrntt: Yeah, that'd add a little too much noise I think.

glbrntt (approved):
Thank you!

PR #191, opened 2026-07-29
Suppress fcntl errors on the listener's accept path

glbrntt:
Motivation:

On Darwin, 'fcntl(2)' on a just-accepted socket can fail with EINVAL, which NIO surfaces as a 'NIOFcntlFailedError'. NIO closes the socket directly and fires the error down the pipeline for informational purposes.

However, when wrapped in a NIOAsyncChannel channel the read loop throws this error which causes the server to shutdown.

Note that testing this reliably is difficult: the error type is public but its init is not so we cannot synthesize the failure. Hitting this is possible in release builds after serveral thousand iterations but much less so in debug builds.

Modifications:

- Ignore the error on Darwin

Result:

Fewer errors


gjcairo (commented):
This LGTM, but I wonder if we should worry about performance by adding another channel handler to the pipeline. Should we add this logic to an existing one instead?

glbrntt:
> This LGTM, but I wonder if we should worry about performance by adding another channel handler to the pipeline. Should we add this logic to an existing one instead?

It's in the listener channel so is only invoked on new connections: it won't register in any perf measurements.

gjcairo (approved):

PR #193, opened 2026-08-03
Add vsock peer credentials on Posix.Context

VictorDebray:
Surfaces the peer's VSOCK context ID (CID) to server handlers as`HTTP2ServerTransport.Posix.Context.virtualSocketCredentials`, and makes a vsock server report the address it actually bound to.

`NIOCore.SocketAddress` has no vsock representation, so `channel.remoteAddress` is always nil for a vsock channel and the CID comes from NIOPosix's `remoteVsockAddress` option instead.
The option reads the peer with `getpeername` and validates the address family, so it's rejected for anything which isn't a vsock socket.

`listeningAddress` had the same problem from the other end: a vsock listening channel has no `localAddress`, so the transport echoed back the configured address, `Port/any` included. The factory now reads the bound address with a `localVsockAddress` option and reports that, so a wildcard bind resolves to the port the kernel picked.

Depends on:
- apple/swift-nio#3689
- the `swift-nio` floor in Package.swift needs raising to the release which ships them
- on #194, which builds the transport-specific context once per connection rather than once per RPC, since reading an option per RPC to identify the peer isn't reasonable.


Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
```diff
@@ -297,6 +301,19 @@ extension HTTP2ServerTransport {
...
+/// Reads the peer's VSOCK context ID (CID) for `channel`.
+///
+/// `channel.remoteAddress` cannot report a vsock peer, because `NIOCore.SocketAddress` has no vsock
+/// representation. NIOPosix exposes the peer address through the `remoteVsockAddress` channel
+/// option instead, which reads it from `getpeername` and validates the address family.
+///
+/// Returns `nil` for any channel that is not a connected vsock channel: the option is rejected for
+/// other address families and is unsupported on non-POSIX channel types.
+@available(gRPCSwiftNIOTransport 2.0, *)
+private func getPeerCID(from channel: any Channel) async -> UInt32? {
```
glbrntt: This is a one-liner only ever called in one place: let's cut out the indirection.

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
```diff
@@ -308,10 +325,27 @@ extension HTTP2ServerTransport.Posix {
...
+
     public init() {
     }
   }
 
+  /// The peer's VSOCK context ID for a virtual-socket connection, taken from
+  /// the connection's peer address. Treat it as the identifier the
+  /// hypervisor/host vsock plumbing reported for the connecting guest.
+  public struct VsockCredentials: Sendable, Equatable {
+    public let cid: UInt32
```
glbrntt: Please use `var` here: `let` doesn't buy you anything.

```swift
var creds = VsockCredentials(cid: 42)
creds.cid += 1 // can't do this
creds = VsockCredentials(cid: cred.cid + 1) // can do this
```

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
```diff
@@ -308,10 +325,27 @@ extension HTTP2ServerTransport.Posix {
...
+    public var vsockCredentials: VsockCredentials?
+
     public init() {
     }
   }
 
+  /// The peer's VSOCK context ID for a virtual-socket connection, taken from
+  /// the connection's peer address. Treat it as the identifier the
+  /// hypervisor/host vsock plumbing reported for the connecting guest.
+  public struct VsockCredentials: Sendable, Equatable {
```
glbrntt: This should be `Hashable`

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
glbrntt: gRPC has a VsockAddress type already, can we use it's `ContextID` type here instead of the raw `UInt32`?

glbrntt (changes requested):
Thanks, we'll need a test for this as well.

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
glbrntt: This is also missing an `@available` annotation, it should be 2.10.

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
```diff
@@ -308,10 +325,27 @@ extension HTTP2ServerTransport.Posix {
...
+    /// time.
+    public var vsockCredentials: VsockCredentials?
+
     public init() {
     }
   }
 
+  /// The peer's VSOCK context ID for a virtual-socket connection, taken from
+  /// the connection's peer address. Treat it as the identifier the
+  /// hypervisor/host vsock plumbing reported for the connecting guest.
```
glbrntt: I think the second sentence is redundant, couldn't this be the other way around too?

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
```diff
@@ -308,10 +320,28 @@ extension HTTP2ServerTransport.Posix {
...
+    public var vsockCredentials: VsockCredentials?
+
     public init() {
     }
   }
 
+  /// The peer's VSOCK context ID for a virtual-socket connection, taken from the connection's
+  /// peer address.
+  @available(gRPCSwiftNIOTransport 2.10, *)
+  public struct VsockCredentials: Hashable, Sendable {
```
glbrntt: This can just be in the core module, I think. It's also not specific to the server transport so should be in its own file at the top level.

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
```diff
@@ -308,10 +320,28 @@ extension HTTP2ServerTransport.Posix {
...
+    public var cid: GRPCNIOTransportCore.SocketAddress.VirtualSocket.ContextID
+
+    public init(cid: GRPCNIOTransportCore.SocketAddress.VirtualSocket.ContextID) {
```
glbrntt: Let's spell this out as `contextID` like the `VirtualSocket` address does.

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
glbrntt: Given we spell out the `VirtualSocket` for the address type we should do the same here: `VirtualSocketCredentials`

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
```diff
@@ -273,6 +273,18 @@ extension HTTP2ServerTransport {
           }
         } catch {}
 
+        // `channel.remoteAddress` can't report a vsock peer because NIO's `SocketAddress` has no
+        // vsock representation, so read the peer address from the channel option instead. It's
+        // rejected for other address families and unsupported on non-POSIX channels, both of which
+        // leave `vsockCredentials` nil.
+        if let vsockAddress = try? await channel.getOption(.remoteVsockAddress).get() {
```
glbrntt: Let's _not_ do this unconditionally: we shouldn't pay the cost of attempting to look up a vsock address if we're not using vsock.

Tests/GRPCNIOTransportHTTP2Tests/VsockCredentialsTests.swift
```diff
@@ -0,0 +1,159 @@
...
+      }
+
+      let control = ControlClient(wrapping: client)
+      let input = ControlInput.with { $0.numberOfMessages = 1 }
+      try await control.unary(request: ClientRequest(message: input)) { response in
+        _ = try response.message
+      }
+
+      server.beginGracefulShutdown()
+      client.beginGracefulShutdown()
```
glbrntt: Just use `withGRPCServer` and `withGRPCClient` here, no need to manually manage each set of resources.

glbrntt: You can also use the `HelloWorldService` here whose implementation can be hooked on init, so the logic of each test can effectively live inside that and be driven by the client making a request. That removes the need for the recorder altogether.

glbrntt (changes requested):

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
VictorDebray: [Address review feedback on vsock peer credentials](https://github.com/grpc/grpc-swift-nio-transport/pull/193/commits/d512d42ccb7d9b38cfc321e9143e6e86c7ad3360)

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
VictorDebray: [Address review feedback on vsock peer credentials](https://github.com/grpc/grpc-swift-nio-transport/pull/193/commits/d512d42ccb7d9b38cfc321e9143e6e86c7ad3360)

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
VictorDebray: [Address review feedback on vsock peer credentials](https://github.com/grpc/grpc-swift-nio-transport/pull/193/commits/d512d42ccb7d9b38cfc321e9143e6e86c7ad3360)

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
VictorDebray: [Address review feedback on vsock peer credentials](https://github.com/grpc/grpc-swift-nio-transport/pull/193/commits/d512d42ccb7d9b38cfc321e9143e6e86c7ad3360)

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
VictorDebray: [Address review feedback on vsock peer credentials](https://github.com/grpc/grpc-swift-nio-transport/pull/193/commits/d512d42ccb7d9b38cfc321e9143e6e86c7ad3360)

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
VictorDebray: [Address review feedback on vsock peer credentials](https://github.com/grpc/grpc-swift-nio-transport/pull/193/commits/d512d42ccb7d9b38cfc321e9143e6e86c7ad3360)

Tests/GRPCNIOTransportHTTP2Tests/VsockCredentialsTests.swift
VictorDebray: [Address review feedback on vsock peer credentials](https://github.com/grpc/grpc-swift-nio-transport/pull/193/commits/6107a5e347af1f69802cc267f0ba4128165c0a66)

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
VictorDebray: [Address review feedback on vsock peer credentials](https://github.com/grpc/grpc-swift-nio-transport/pull/193/commits/6107a5e347af1f69802cc267f0ba4128165c0a66)

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
VictorDebray: [Address review feedback on vsock peer credentials](https://github.com/grpc/grpc-swift-nio-transport/pull/193/commits/6107a5e347af1f69802cc267f0ba4128165c0a66)

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
VictorDebray: [Address review feedback on vsock peer credentials](https://github.com/grpc/grpc-swift-nio-transport/pull/193/commits/6107a5e347af1f69802cc267f0ba4128165c0a66)

VictorDebray:
I updated the PR with your review. I have verified the changes work in our project too end to end.

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
```diff
@@ -273,6 +273,20 @@ extension HTTP2ServerTransport {
           }
         } catch {}
 
+        // NIO's `SocketAddress` has no vsock representation, which has two consequences here: a
+        // vsock channel is exactly the case where `localAddress` is absent, so that's the cheap
+        // test for whether it's worth asking at all; and the peer's address has to come from the
+        // channel option rather than from `remoteAddress`.
+        if channel.localAddress == nil,
```
glbrntt: Just switch on the `address` passed to the `init`, we don't need to use indirect logic here.

Tests/GRPCNIOTransportHTTP2Tests/VirtualSocketCredentialsTests.swift
```diff
@@ -0,0 +1,133 @@
...
+  /// and the reported context ID is `VMADDR_CID_LOCAL`.
+  @Test(
+    "Peer credentials describe the connecting context over a virtual socket",
+    .enabled(if: vsockLoopbackAvailable(), "Vsock loopback is unavailable")
+  )
+  @available(gRPCSwiftNIOTransport 2.10, *)
+  func credentialsOverVirtualSocket() async throws {
+    // `VMADDR_CID_LOCAL`, which routes to the host that generated the packets. gRPC's `ContextID`
+    // exposes `any`, `hypervisor` and `host`, but not `local`.
+    let localContextID = GRPCNIOTransportCore.SocketAddress.VirtualSocket.ContextID(rawValue: 1)
```
glbrntt: Why not just add `local`?

Tests/GRPCNIOTransportHTTP2Tests/VirtualSocketCredentialsTests.swift
```diff
@@ -0,0 +1,133 @@
...
+    // NIO's `SocketAddress` can't represent a vsock address, so the server's listening address is
+    // whatever it was configured with and binding `.any` would leave the client with no port to
+    // connect to. That forces a fixed port, so use one unlikely to be in use.
```
glbrntt: If the functionality is being added to NIO to be able to recover the remote address, should it not also be matched with the functionality to recover the local vsock address?

Run tests enough and port collisions _will_ happen.

glbrntt (changes requested):

Tests/GRPCNIOTransportHTTP2Tests/VirtualSocketCredentialsTests.swift
VictorDebray: Done in [Use ContextID.local in the vsock credentials test](https://github.com/grpc/grpc-swift-nio-transport/pull/193/commits/406925102fb96fca3b5c07ae4923c97de3b2a50f). `ContextID.local` already exist. The comment was wrong, so I removed it.

Tests/GRPCNIOTransportHTTP2Tests/VirtualSocketCredentialsTests.swift
VictorDebray: Fixed in [Report the address a vsock listener bound to](https://github.com/grpc/grpc-swift-nio-transport/pull/193/commits/b358cc43f9cb872082de8832358795d8c42d1229).
I've added a `localVsockAddress` channel option to the `swift-nio` PR https://github.com/apple/swift-nio/pull/3689 along the `remoteVsockAddress`.
The listener factory reads the bound address off the listening channel with it and `listeningAddress` reports that, so a `vsock` server describes where it actually is instead of echoing the wildcard it was configured with. The test binds `Port/any` and connects to whatever comes back, so there's no fixed port left to collide on.

Reading the option is also how the factory learns that it *is* a vsock listener, which is the only way to report the address of one gRPC was handed as a descriptor.

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
VictorDebray: [Ask whether a connection is a vsock connection rather than inferring it](https://github.com/grpc/grpc-swift-nio-transport/pull/193/commits/b1a16b599606b3217df2391dcc4cd2a9abc75217), though I chose not  to switch on the `init` address in the end. Let me know what you think!

I found that gRPC can be handed an already-bound listening descriptor, and there's no address to switch on there, so a vsock listener created that way would have silently stopped reporting credentials.

What I'm doing instead is not not infer on anything: the channel option validates the socket's address family and is rejected for anything which isn't a vsock socket.

The transport-specific context closure runs once per *RPC* rather than once per connection, so on mTLS servers the peer's certificates are re-read and re-parsed from DER on every request today. Computing the transport-specific context once per connection fixes that, and with it in an unconditional read costs one rejected option read per connection.
Depends on #194 to go in first.

VictorDebray:
Two things this depends on, so it can't go green on its own yet:

- swift-nio. It needs `remoteVsockAddress` (apple/swift-nio#3689) and `localVsockAddress`, which I've added to the same branch.
- #194

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
```diff
@@ -156,12 +178,34 @@ extension HTTP2ServerTransport {
...
+            .vsock(
+              contextID: GRPCNIOTransportCore.SocketAddress.VirtualSocket.ContextID(
+                rawValue: bound.cid.rawValue
+              ),
+              port: GRPCNIOTransportCore.SocketAddress.VirtualSocket.Port(
+                rawValue: bound.port.rawValue
+              )
+            )
+          )
+        }
```
glbrntt, line 200: I don't think this approach is quite right. I think the `ListenerConfigurator` should be extended with API that allows listener factories to set a bound address.

The posix transport would then be able to set that appropriately from the underlying transport.

Sources/GRPCNIOTransportHTTP2Posix/HTTP2ServerTransport+Posix.swift
```diff
@@ -281,6 +336,21 @@ extension HTTP2ServerTransport {
...
         } catch {}
 
+        // NIO's `SocketAddress` has no vsock representation, so the peer's address has to come from
+        // the channel option rather than from `remoteAddress`.
+        //
+        // The option validates the socket's address family, so asking is how we find out whether
+        // this is a vsock connection: it's rejected for anything else. This runs once per
+        // connection, so a connection which isn't a vsock connection pays for one rejected read
+        // rather than one per RPC.
+        if let vsockAddress = try? await channel.getOption(.remoteVsockAddress).get() {
```
glbrntt, line 346: We should only be doing this if we know we have a vsock connection.

glbrntt (changes requested):

VictorDebray:
It now switches on the address passed to the init which is the same test the listener uses shared as `Address.mayBeVirtualSocket`. So now TCP and UDS servers never read the option.

The exception is a listener handed to gRPC as a descriptor. It's bound before the transport sees it, so there's no address to switch on and we ask anyway: one `remoteVsockAddress` read per accepted connection that are rejected unless the socket really is a vsock socket.

We could not do that check if we store the answer. The factory already resolves the family at bind. It reads localVsockAddress to report the bound address, and accepted sockets inherit the listener's family.
But that would mean carrying per-transport state you asked me to drop in favour of setBoundAddress.

What do you suggest doing? Keep the speculative read for descriptors or go the state route ?

glbrntt (changes requested):
Let's simplify this, this PR has been open for 7 weeks now and it doesn't feel a whole lot closer to being merged.

Let's divide this into two PRs:
1. Adding vsock credentials to the context when vsock is known to be used (i.e. pre-bound socket case, that can always be added later if a user needs it). This should verify that when vsock is used on a supported platform that the context has a non-nil vsock credentials object on it. It does not need to verify anything else.
2. Add the listening address to the server. I think this is better done by adding another requirement the `ListenerFactory`, something like `func listeningAddress(of channel: any Channel) async throws -> SocketAddress?`. It'll need a default implementation returning `channel.localAddress.map { SocketAddress($0) }` though.

VictorDebray:
Great idea!
Here they are:
- [Report the listening address from the listener factory](https://github.com/grpc/grpc-swift-nio-transport/pull/197#top)
- [Add vsock peer credentials on Posix.Context](https://github.com/grpc/grpc-swift-nio-transport/pull/198#top)
I'm closing this PR in favor of these.

PR #194, opened 2026-08-20
Compute the transport-specific context once per connection

VictorDebray:
`Custom` built the transport-specific server context inside `handleStream`, so the closure ran for every RPC on a connection even though everything it can report describes the connection. On the Posix transport that meant re-reading the peer's certificates and re-parsing them from DER on each request.

The context is now built once per connection and shared by every stream on it.

It's built when the first stream arrives rather than alongside `remotePeer` and `localPeer` because it can need more than an active channel: the Posix transport reads the peer's TLS certificates, which aren't available until the handshake has completed. A stream arriving implies an established connection, which is what made building it per stream correct.

`handleStream` no longer needs the connection channel, so it takes the context to use instead.

The test drives two RPCs over a single connection and asserts the context is built once and seen by both.

Needed for #193, which reads a vsock channel option to report the peer's context ID and can only justify that once per connection.


Sources/GRPCNIOTransportCore/Server/CustomTransport.swift
```diff
@@ -324,15 +324,33 @@ extension HTTP2ServerTransport {
...
+            // stream arriving implies an established connection, which is what made computing this
+            // per stream correct.
+            //
+            // The work happens in its own task so that waiting for it doesn't stop this loop from
+            // accepting streams: each stream waits for the same result instead.
+            var transportSpecific: Task<any ServerContext.TransportSpecific, Never>?
+
             for try await (stream, descriptor) in multiplexer.inbound {
+              if transportSpecific == nil, let makeContext = self.transportSpecificContext {
+                transportSpecific = Task { await makeContext(connection.channel) }
```
glbrntt, line 342: There's no need for an unstructured task here and the comment is wrong about this blocking streams being accepted.

Sources/GRPCNIOTransportCore/Server/CustomTransport.swift
```diff
@@ -324,15 +324,33 @@ extension HTTP2ServerTransport {
...
+            // at most once and share it with every stream.
+            //
+            // It's computed when the first stream arrives rather than alongside the peer addresses
+            // above because it can need more than an active channel: the Posix transport reads the
+            // peer's TLS certificates, which aren't available until the handshake has completed. A
+            // stream arriving implies an established connection, which is what made computing this
+            // per stream correct.
+            //
+            // The work happens in its own task so that waiting for it doesn't stop this loop from
+            // accepting streams: each stream waits for the same result instead.
```
glbrntt, line 337: This is incorrect: it does not stop streams from being accepted. They have already been accepted by the connection.

Sources/GRPCNIOTransportCore/Server/CustomTransport.swift
```diff
@@ -324,15 +324,33 @@ extension HTTP2ServerTransport {
           }
 
           do {
+            // The transport-specific context describes the connection, not the RPC, so compute it
+            // at most once and share it with every stream.
+            //
```
glbrntt, line 329: unnecessary context

Sources/GRPCNIOTransportCore/Server/CustomTransport.swift
```diff
@@ -324,15 +324,33 @@ extension HTTP2ServerTransport {
...
+            // It's computed when the first stream arrives rather than alongside the peer addresses
+            // above because it can need more than an active channel: the Posix transport reads the
+            // peer's TLS certificates, which aren't available until the handshake has completed. A
+            // stream arriving implies an established connection, which is what made computing this
+            // per stream correct.
```
glbrntt, line 334: This is a lot of text to say very little.

glbrntt (changes requested):

Sources/GRPCNIOTransportCore/Server/CustomTransport.swift
VictorDebray: [Compute the transport-specific context without an unstructured task](https://github.com/grpc/grpc-swift-nio-transport/pull/194/commits/a9e4b79c8418758f8006724dbbb3797920ba7afd)
Fixed it up thanks

Sources/GRPCNIOTransportCore/Server/CustomTransport.swift
VictorDebray: [Compute the transport-specific context without an unstructured task](https://github.com/grpc/grpc-swift-nio-transport/pull/194/commits/a9e4b79c8418758f8006724dbbb3797920ba7afd)

Sources/GRPCNIOTransportCore/Server/CustomTransport.swift
VictorDebray: removed

Sources/GRPCNIOTransportCore/Server/CustomTransport.swift
VictorDebray: removed

Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportNIOPosixTests.swift
```diff
@@ -637,6 +639,84 @@ final class HTTP2TransportNIOPosixTests: XCTestCase {
       }
     }
   }
+
+  /// The transport-specific context describes the connection, so it should be built once per
+  /// connection however many RPCs are made over it, each RPC on a connection should see the
+  /// context built for it, and a second connection should get its own.
+  @available(gRPCSwiftNIOTransport 2.6, *)
```
glbrntt:
```suggestion
  @available(gRPCSwiftNIOTransport 2.6, *)
```

Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportNIOPosixTests.swift
```diff
@@ -637,6 +639,84 @@ final class HTTP2TransportNIOPosixTests: XCTestCase {
...
+        return CountingTransportSpecific(id: id)
+      }
+    )
+
+    try await withGRPCServer(transport: transport, services: [service]) { _ in
+      let address = await transport.listeningAddress
+      let ipv4Address = try XCTUnwrap(address?.ipv4)
+
+      // Each client is one connection: it has a single endpoint to pick from, so the RPCs it
+      // makes all go over the same one.
```
glbrntt:
```suggestion
```

Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportNIOPosixTests.swift
```diff
@@ -637,6 +639,84 @@ final class HTTP2TransportNIOPosixTests: XCTestCase {
...
+      XCTAssertEqual(contextsMade.withLock { $0 }, 2)
+
+      let seen = contextsSeen.withLock { $0 }
+      let first = try XCTUnwrap(seen["first"])
+      let second = try XCTUnwrap(seen["second"])
+      let third = try XCTUnwrap(seen["third"])
+
+      // Built once, but handed to both RPCs on the connection.
+      XCTAssertEqual(first, second)
+      // The second connection got its own: the context isn't cached beyond the connection.
```
glbrntt:
```suggestion
      XCTAssertEqual(first, second)
```

Tests/GRPCNIOTransportHTTP2Tests/HTTP2TransportNIOPosixTests.swift
```diff
@@ -637,6 +639,84 @@ final class HTTP2TransportNIOPosixTests: XCTestCase {
...
+      XCTAssertNotEqual(first, third)
+    }
+  }
+}
+
+/// A transport-specific context which exists only to be identified.
+@available(gRPCSwiftNIOTransport 2.6, *)
+private struct CountingTransportSpecific: ServerContext.TransportSpecific {
+  /// Distinguishes this context from those built for other connections.
```
glbrntt:
```suggestion
@available(gRPCSwiftNIOTransport 2.6, *)
private struct CountingTransportSpecific: ServerContext.TransportSpecific {
```

Sources/GRPCNIOTransportCore/Server/CustomTransport.swift
```diff
@@ -326,15 +326,26 @@ extension HTTP2ServerTransport {
           }
 
           do {
+            // Computed when the first stream arrives, not alongside the peer addresses above: the
+            // Posix transport reads the peer's TLS certificates, which aren't available until the
+            // handshake has completed.
```
glbrntt:
```suggestion
            // Computed on the first stream so that any TLS handshake has completed.
```

glbrntt (approved):

